PR #21, opened 2017-09-14
feat: progress flag for CloudVolume, Storage

### LINKED ISSUE #20: Replace queue.join()
ThreadedQueue uses queue.join() which blocks until the entire queue is done. This prevents on demand error checking and status updates _a la_ tqdm. This can probably be replaced with a while loop and sleep that will process the queue in smaller batches before checking up on things.

william-silversmith:
```
CloudVolume(..., progress=True) # Shows progress bars
Storage(..., progress=True) # shows progress bars
```

ThreadedQueue now checks every 0.1 seconds for exceptions and updates
a progress indicator. Storage now takes advantage of this for
multithreaded uploads and downloads. CloudVolume now supports a
vol.progress flag that switches tqdm output on and off, default False.

Updated some comments for Storage as well.

Resolves #20

### DIFF
--- a/cloudvolume/cloudvolume.py
+++ b/cloudvolume/cloudvolume.py
@@ -694,7 +694,7 @@ class CloudVolume(object):
     if str(self.dtype) != str(img.dtype):
       raise ValueError('The uploaded image data type must match the volume data type. volume: {}, image: {}'.format(self.dtype, img.dtype))
 
-    iterator = tqdm(self._generate_chunks(img, offset), disable=(not self.progress), desc='rechunking image')
+    iterator = tqdm(self._generate_chunks(img, offset), desc='Rechunking image', disable=(not self.progress))
 
     uploads = []
     for imgchunk, spt, ept in iterator:
--- a/cloudvolume/cloudvolume.py
+++ b/cloudvolume/cloudvolume.py
@@ -67,8 +67,11 @@ class CloudVolume(object):
         False: Throw an error
     info: (dict) in lieu of fetching a neuroglancer info file, use this provided one.
             This is useful when creating new datasets.
+    progress: (bool) Show tqdm progress bars.
   """
-  def __init__(self, cloudpath, mip=0, bounded=True, fill_missing=False, info=None):
+  def __init__(self, cloudpath, mip=0, bounded=True, fill_missing=False, 
+      progress=False, info=None):
+
     super(self.__class__, self).__init__()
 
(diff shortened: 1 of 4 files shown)


william-silversmith:
Something curious about this. The main thread sleepy spin lock I implemented seems to cause an increase of about 1-1.5 sec over a simple `self._queue.join()` when downloading test_v0. 

That is about 16-17s for old implementation vs 17.3-18.2 for this version. Increasing the sleep time doesn't seem to make much of a difference.

william-silversmith:
Some further testing showed that this effect is within the margin of variation. A more controlled experiment is needed to determine if it's real.

william-silversmith:
@davidbuniat ran into an issue with CloudVolume where he was running CloudVolume inside of a multithreaded environment. He ran into a multithreading issues with `tqdm`. I wonder if this would fix that with `progress=False`

davidbuniat:
my issue connected with tqdm on this branch seems resolved

cloudvolume/cloudvolume.py
```diff
@@ -640,8 +652,10 @@ def upload_image(self, img, offset):
     if str(self.dtype) != str(img.dtype):
       raise ValueError('The uploaded image data type must match the volume data type. volume: {}, image: {}'.format(self.dtype, img.dtype))
 
+    iterator = tqdm(self._generate_chunks(img, offset), disable=(not self.progress), desc='rechunking image')
```
wongwill86: For consistence, you should match the order of the args (in `filteiter` you did `desc` then `disable`).
also capital R in `rechunking image`

cloudvolume/threaded_queue.py
```diff
@@ -175,30 +179,62 @@ def _consume_queue_execution(self, fn):
...
+      desc = progress
+
+    last = self._inserted
+    with tqdm(total=self._inserted, disable=(not progress), desc=desc) as pbar:
+      # Allow queue to consume, but check up on
+      # progress and errors every tenth of a second
+      while not self._queue.empty():
+        size = self._queue.qsize()
+        delta = last - size
+        if delta != 0: # We should crash on negative numbers
```
wongwill86: how is it possible to get negative numbers?

william-silversmith: Programming errors when you are changing the code.

william-silversmith: Also if someone interferes with self._inserted

PR #25, opened 2017-09-21
fix: python3 compatibility

william-silversmith:
Reading a JSON file but not encoding it as utf-8 means it remains
a bytes object.


ranlu:
Looks good to me, tried on my workstation and it works with python3 without a glitch now.

PR #30, opened 2017-09-23
feat: add support for intermediate paths in URIs

### LINKED ISSUE #28: Add support for dataset_path to CV
This way you don't have to worry about making file:// protocols conform to the bucket/dataset/path from the root.

william-silversmith:
Resolves #28 

Add support for things like:

CloudVolume('file:///usr/home/username/BUCKET/DATASET/LAYER/')

### DIFF
--- a/cloudvolume/cloudvolume.py
+++ b/cloudvolume/cloudvolume.py
@@ -28,7 +28,7 @@ else:
 __all__ = [ 'CloudVolume', 'EmptyVolumeException' ]
 
 ExtractedPath = namedtuple('ExtractedPath', 
-  ('protocol','bucket_name', 'dataset_name','layer_name')
+  ('protocol', 'intermediate_path', 'bucket', 'dataset','layer')
 )
 
 class EmptyVolumeException(Exception):
@@ -71,21 +71,14 @@ class CloudVolume(object):
   def __init__(self, cloudpath, mip=0, bounded=True, fill_missing=False, info=None):
     super(self.__class__, self).__init__()
 
-    extract = CloudVolume.extract_path(cloudpath)
-
-    self._protocol = extract.protocol
-    self._bucket = extract.bucket_name
-
-    # You can access these two with properties
-    self._dataset_name = extract.dataset_name
-    self._layer = extract.layer_name
+    self.path = CloudVolume.extract_path(cloudpath)
 
     self.mip = mip
     self.bounded = bounded
     self.fill_missing = fill_missing
 
     self._storage = None
-    if self._protocol != 'boss':
+    if self.path.protocol != 'boss':
       self._storage = Storage(self.layer_cloudpath, n_threads=0)
 
     if info is None:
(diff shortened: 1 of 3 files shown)


cloudvolume/cloudvolume.py
```diff
@@ -143,11 +136,22 @@ def create_new_info(cls, num_channels, layer_type, data_type, encoding, resoluti
   @classmethod
   def extract_path(cls, cloudpath):
     """cloudpath: e.g. gs://neuroglancer/DATASET/LAYER/info or s3://..."""
-    match = re.match(r'^(gs|file|s3|boss)://(/?[\d\w_\.\-]+)/([\d\w_\.\-]+)/([\d\w_\.\-]+)/?', cloudpath)
-    return ExtractedPath(*match.groups())
+    protocol_re = r'^(gs|file|s3|boss)://'
```
tartavull: Why so many re.match and re.sub?
You can just do it all in one line
https://github.com/seung-lab/neuroglancer/blob/it_queue/python/neuroglancer/pipeline/storage.py#L63

cloudvolume/cloudvolume.py
```diff
@@ -143,11 +136,22 @@ def create_new_info(cls, num_channels, layer_type, data_type, encoding, resoluti
...
+    tail_re = r'(/?[\d\w_\.\-]+)/([\d\w_\.\-]+)/([\d\w_\.\-]+)(?:/info|/provenance|/)?$'
```
tartavull: Accepting an info there is sort of problematic, what if my layer name is `info`?
Interpreting "common" mistake vs making the definition simpler.
I would maybe check and return a message like:
"Is your layer actually call info? otherwise remove info from here"

cloudvolume/cloudvolume.py
```diff
@@ -211,15 +215,15 @@ def commitInfo(self):
     return self.commit_info()
 
   def commit_info(self):
-    if self._protocol == 'boss':
+    if self.path.protocol == 'boss':
```
tartavull: All this line is some sort of silent failure ,right?
info file cannot be changed for boss, but when we are trying to save the change we do nothing
the user thinks she actually commited the changes.

This applies to refresh_provenance and others places

cloudvolume/cloudvolume.py
```diff
@@ -238,31 +242,19 @@ def refresh_provenance(self):
...
   @property
   def dataset_name(self):
-    return self._dataset_name
-
-  @dataset_name.setter
-  def dataset_name(self, name):
-    if name != self._dataset_name:
-      self._dataset_name = name
-      self.refresh_info()
+    return self.path.dataset
```
tartavull: How much time are we saving the user on this?
he can just use the path to get to dataset himself.

cloudvolume/cloudvolume.py
```diff
@@ -277,7 +269,7 @@ def mip_scale(self, mip):
 
   @property
   def base_cloudpath(self):
-    return "{}://{}/{}/".format(self._protocol, self._bucket, self.dataset_name)
+    return self.path.protocol + "://" + os.path.join(self.path.intermediate_path, self.path.bucket, self.dataset_name)
```
tartavull: I don't find the name base_cloudpath intuitive.
why not just path.join()?

tartavull: oh I see, this returns the path to the dataset

cloudvolume/cloudvolume.py
william-silversmith: I found that line hard to understand and maybe error prone. I figured processing the prefix, the suffix, and then the part in the middle I was more likely to get right.

cloudvolume/cloudvolume.py
william-silversmith: When I originally wrote this, it was the fastest way to get `boss` working without needing to rewrite a bunch of tasks that call `commit_info`. The right way to do it is probably to actually try committing the info file to BOSS. BOSS also doesn't support provenance files per se. I think we're going to address this in a refactoring that separates cutout services from Precomputed based key-value stores.

cloudvolume/cloudvolume.py
william-silversmith: We discussed and think that:
1) Neuroglancer doesn't support trailing info/provenance so we'd be training people in two different ways. Furthermore, even if we change our version of neuroglancer, there are many that won't be changed.
2) We can identify problems and issue corrective advice if the info fetch fails and the layer is named info.

william-silversmith:
Per conversation with Nacho, g2g.

PR #38, opened 2017-10-11
feat: CloudVolume supports intermediate paths

william-silversmith:
This required refactoring the path extraction logic out of
CloudVolume and Storage.

### DIFF
--- a/cloudvolume/cloudvolume.py
+++ b/cloudvolume/cloudvolume.py
@@ -1,6 +1,5 @@
 from __future__ import print_function
 
-from collections import namedtuple
 import json
 import json5
 import os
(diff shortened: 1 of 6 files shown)


jingpengw:
@william-silversmith just curious, what does a intermediate path mean? any example?

william-silversmith:
I wanted ppl to be able to, as Will Wong suggested, set cache=True, then later on just point their cloud volume at the cache. e.g. 

```
vol = CloudVolume('gs://neuroglancer/test_v0/image', cache=True)
.... downloads parts or all of volume ...
... later...
vol = CloudVolume('file:///Users/wms/.cloudvolume/cache/gs/neuroglancer/test_v0/image')
```

The intermediate path is: `wms/.cloudvolume/cache/gs/neuroglancer` in this case.

PR #48, opened 2017-10-23
feat: save meshes from a url (save_mesh_from_url)

william-silversmith:


william-silversmith:
The equivalences algorithm is wrong, it only does a depth 1 search of the tree.

william-silversmith:
Not worth the effort at the moment. We can revisit, but it's more likely neuroglancer will get the ability to download meshes directly.

Issue #49, opened 2017-10-24
threaded_queue.py far too sleepy


nkemnitz:
I am using cloud-volume in Julia for time-sensitive loading of tiny cutouts. The chunks are already in `.cloudvolume/cache`, so I expect loading times around 50ms.
In reality, loading the two compressed files from cache takes around 1.02 seconds with this script :

```python
from cloudvolume import CloudVolume
vol = CloudVolume('gs://neuroglancer/pinky40_v11/watershed', bounded = False, cache = True)
image = vol[26045:26077, 20874:20906, 795:799] # download and send to cache, should be somewhat slow
image = vol[26045:26077, 20874:20906, 795:799] # load from cache, should be very fast
```

Running cProfile for the 4th line only tells me that the additional second is entirely spent [sleeping in the threaded_queue](https://github.com/seung-lab/cloud-volume/blob/master/cloudvolume/threaded_queue.py#L224)
```
         2882 function calls in 1.038 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       10    1.002    0.100    1.002    0.100 {time.sleep}
      129    0.017    0.000    0.017    0.000 {method 'acquire' of 'thread.lock' objects}
        1    0.004    0.004    1.037    1.037 cloudvolume.py:803(_cutout)

                                    ... *snip* ...
```

Do you have a quick solution/hack how I can get reasonable times for this scenario using CloudVolume.jl? Maybe force cloud-volume to use a single thread and avoid the queue all together? A proper solution would be even better, of course. :)

william-silversmith:
Should be resolved for now. Reopen if you run into another issue.

PR #50, opened 2017-10-24
Optimize Sleep Patterns

william-silversmith:
In response to #49


william-silversmith:
Per discussion, speed up does resolve current issue. If we run into another bottleneck, we should take another look at this architecture.

PR #51, opened 2017-10-24
feat: added cache-control options to Storage

### LINKED ISSUE #31: Reduce Cache-Control max-age for info File
Interferes with seeing results.

william-silversmith:
Set cache-control to no-cache for info and provenance files

Resolves #31


william-silversmith:
It would be nice if setting cache-control worked for cached local files too. In theory might be possible, but it'll make things more complex since file metadata doesn't natively support this.

PR #61, opened 2017-11-13
feat: better error message for missing mesh manifests

william-silversmith:
Now prints when using vol.save_mesh(...):
Segment ID(s) $SEGIDS are missing corresponding mesh manifests.
Aborted.

### DIFF
--- a/cloudvolume/cloudvolume.py
+++ b/cloudvolume/cloudvolume.py
@@ -1029,13 +1029,17 @@ class CloudVolume(object):
     
       yield chunkimg, spt, ept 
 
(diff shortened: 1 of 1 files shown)


william-silversmith:
There are a few ways to speed this up. For instance, you can pull the entire manifest file and check if it 404s instead of testing for existence and reuse them later. This would also allow you to avoid pulling the manifest files one by one and thus call `stor.get_files` on all fragments at once. 

However, this version is relatively clearly written so I'll defer that work until we run into a performance bottleneck.

PR #69, opened 2018-02-24
Concurrent painting w/ Download (Save image size worth of memory)

william-silversmith:
Downloading:

parallel

Blue = old algorithm on chunked data
Red = New algorithm on chunked data
Black = New Algorithm on single chunk download (but the old algorithm looks the same too)


william-silversmith:
For Upload:

Before:
before

After:
after

PR #86, opened 2018-03-23
feat: preserve option for fush_cache + flush_cache_region

### LINKED ISSUE #84: Function to Delete Subsections of Cache
Requested by @jingpengw 

See discussion in #86

william-silversmith:
Resolves #84

### DIFF
--- a/cloudvolume/cacheservice.py
+++ b/cloudvolume/cacheservice.py
@@ -1,0 +1,244 @@
+import os
+import shutil
+
+from .provenance import DataLayerProvenance
+from .storage import SimpleStorage, Storage
+
+from .lib import Bbox, colorize, jsonify, mkdir, toabs, Vec, generate_slices
+from .secrets import CLOUD_VOLUME_DIR
+
+def warn(text):
+  print(colorize('yellow', text))
+
+class CacheService(object):
+
+  def __init__(self, path_or_bool, vol):
+    self.vol = vol
+    self.enabled = path_or_bool # on/off or path (on)
+
+    self.initialize()
+
+  def initialize(self):
+    if not self.enabled:
+      return 
+
+    if not os.path.exists(self.path):
+        mkdir(self.path)
+
+    if not os.access(self.path, os.R_OK|os.W_OK):
+      raise IOError('Cache directory needs read/write permission: ' + self.path)
+
+  @property
+  def path(self):
+    if type(self.enabled) is not str:
+      path = self.vol.path
+      return toabs(os.path.join(CLOUD_VOLUME_DIR, 'cache', 
+        path.protocol, path.bucket.replace('/', ''), path.intermediate_path,
+        path.dataset, path.layer
+      ))
+    else:
+      return toabs(self.enabled)
+  
+  def num_files(self, all_mips=False):
+    def size(mip):
+      path_at_mip = os.path.join(self.path, self.vol.mip_key(mip))
+      if not os.path.exists(path_at_mip):
+        return 0
+      return len(os.listdir(path_at_mip))
+
+    if all_mips:
+      sizes = [ 0 ] * (max(list(self.vol.available_mips)) + 1)
+      for i in self.vol.available_mips:
+        sizes[i] = size(i)
+      return sizes
+    else:
+      return size(self.vol.mip)
+
+  def num_bytes(self, all_mips=False):
+    def mip_size(mip):
+      path_at_mip = os.path.join(self.path, self.vol.mip_key(mip))
+      if not os.path.exists(path_at_mip):
+        return 0
+
+      return sum( 
+        ( os.path.getsize(os.path.join(path_at_mip, filename)) for filename in os.listdir(path_at_mip) ) 
+      )
+
+    if all_mips:
+      sizes = [ 0 ] * (max(list(self.vol.available_mips)) + 1)
+      for i in self.vol.available_mips:
+        sizes[i] = mip_size(i)
+      return sizes
+    else:
+      return mip_size(self.vol.mip)
+
+  def flush(self, preserve=None):
+    """
+    Delete the cache for this dataset. Optionally preserve
+    a region. Helpful when working with overlaping volumes.
+
+    Warning: the preserve option is not multi-process safe.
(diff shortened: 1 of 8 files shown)


jingpengw:
thanks, this is quick and helpful!
I have a few questions:
- do we have function to list the cached chunks?
- can we get the total size of current cache (measured by GB/MB)
- do we have function to preserve a specific list of cached chunks
- or have a service to maintain a cache pool with limited cache size or chunk number.

william-silversmith:
I'm torn with respect to how many cache functions I should implement since the underlying implementation is so simple that it might be better to show people how to use it. If I add the code, the implementation of cloudvolume becomes more complex to maintain.

To list cached items, this code suffices:

```
import os
path_at_mip = os.path.join(vol.cache_path, vol.key)
for filename in os.listdir(path_at_mip):
    do what you want
```

To get the total size:
```
import os
num_bytes = 0
path_at_mip = os.path.join(vol.cache_path, vol.key)
for filename in os.listdir(path_at_mip):
    num_bytes += os.path.getsize(os.path.join(path_at_mip, filename))
```

To preserve only a set of cached chunks of your choosing, just use set and set intersection to delete all others.

Understanding the very simple cache is a lot more powerful than lots of custom features. However, I think I can get behind offering cache size utilities. It should be a one liner for a particular data layer. 

The service angle is a little more complicated. You start running into decisions like: 
- Should I enforce cache size whenever a read takes place or only on disk writes? 
- What if the requested cache volume is greater than my limit?
- What if someone has downloaded > 1TB of files? An automatic process may cause unexpected slowdown.
- What if multiple cloud volumes are operating in parallel on the same FS? This is typical on local machines if not in the cloud.
- What default assumptions for a cache size are reasonable? Right now the assumption is your whole free disk space is eligible. This assumption can be valuable for local use where a user is experimenting with a large section of data, but may be less appropriate for an operationalized cloud process.

Thoughts?

william-silversmith:
A slight complication in the above examples is that since on-disk cache files can have a .gz suffix and be compressed, it might make sense to access them through Storage rather than manually.

william-silversmith:
@jingpengw how do you feel about the latest changes? You can now run `vol.cache.num_bytes()` and `vol.cache.num_files()`. They both take an all_mips flag that will cause them to return a list that contains the number for each mip level so you can wrap it in `sum()` if you want a larger count.

cloudvolume/volumecutout.py
```diff
@@ -0,0 +1,92 @@
...
+        img = self[:, :, level, channel ]
+      else:
+        raise NotImplemented
+
+      num_channels = img.shape[2]
+
+      for channel_index in range(num_channels):
+        img2d = img[:, :, channel_index]
+
+        # discovered that downloaded cube is in a weird rotated state.
```
jingpengw: why? it is still a row major and column major problem?

cloudvolume/volumecutout.py
```diff
@@ -0,0 +1,92 @@
...
+
+from .lib import mkdir
+
+class VolumeCutout(np.ndarray):
+
+  def __new__(cls, buf, dataset_name, layer, mip, layer_type, bounds, *args, **kwargs):
+    return super(VolumeCutout, cls).__new__(cls, shape=buf.shape, buffer=np.ascontiguousarray(buf), dtype=buf.dtype)
+
+  def __init__(self, buf, dataset_name, layer, mip, layer_type, bounds, *args, **kwargs):
+    super(VolumeCutout, self).__init__()
```
jingpengw: for python3, just `super()` rather than `super(VolumeCutout, self)` will do the job.

cloudvolume/volumecutout.py
```diff
@@ -0,0 +1,92 @@
+import os
```
jingpengw: what does this special class for?

cloudvolume/volumecutout.py
```diff
@@ -0,0 +1,92 @@
...
+  def save_images(self, axis='z', channel=None, directory=None, image_format='PNG'):
+
+    if directory is None:
+      directory = os.path.join('./saved_images', self.dataset_name, self.layer, str(self.mip), self.bounds.to_filename())
+    
+    mkdir(directory)
+
+    print("Saving to {}".format(directory))
+
+    indexmap = {
```
jingpengw: IntEnum is a better fitting?
https://docs.python.org/3/library/enum.html#intenum

cloudvolume/cacheservice.py
```diff
@@ -0,0 +1,244 @@
...
+
+    if all_mips:
+      sizes = [ 0 ] * (max(list(self.vol.available_mips)) + 1)
+      for i in self.vol.available_mips:
+        sizes[i] = size(i)
+      return sizes
+    else:
+      return size(self.vol.mip)
+
+  def num_bytes(self, all_mips=False):
```
jingpengw: these two functions are very helpful!

cloudvolume/volumecutout.py
william-silversmith: It's what CloudVolume returns from downloads. It's derived from np.ndarray and you can use it the same way. It comes with a few helper items that track where it came from and a way to serialized images to disk for inspection.

cloudvolume/volumecutout.py
william-silversmith: Possibly, but I this is simpler and works in this small situation.

cloudvolume/cacheservice.py
william-silversmith: I'm glad you like them!

cloudvolume/volumecutout.py
william-silversmith: Technically CloudVolume is python2 and python3 compatible. If we want to drop Python2 we should make that decision affirmatively.

cloudvolume/volumecutout.py
william-silversmith: It's something like that. It probably has something to do with row major and column major + an interaction with the PNG format. A mere transpose isn't enough to correct it. I think it might be that PNGs index from the top left while our coordinate system is bottom right.

PR #91, opened 2018-04-02
Changing the docs a bit

pgunn:
Extracted bits of my runner to add into the docs, also changed some wording for easier reading. If it'd be helpful we could instead make an "examples" directory and I could make a friendlier version of my runner in there.


william-silversmith:
Thank you Pat!

PR #92, opened 2018-04-06
feat: Added non-aligned chunk uploads.

william-silversmith:
refactor: moved precomputed upload/download to txrx

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,2 +1,4 @@
+Jingpeng Wu <jingpeng.wu@gmail.com>
+Pat Gunn <pgunn01@gmail.com>
 Thomas Macrina <thomas.macrina@gmail.com>
 William Silversmith <william.silversmith@gmail.com>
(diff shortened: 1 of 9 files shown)


william-silversmith:
- Still need to add automated tests
- Adjust warning message to drop the implication that I used the dumb implementation
- Check benchmarks to make sure nothing changed.
- Double check generate_chunks usage to make sure dropping alignment check is ok.
- Comment on setflag(write=1)
- Add function header comments. 

This refactor paves the way for sensible parallel upload/download. The upload/download functions cannot be part of a class to use multiprocessing.

jingpengw:
@william-silversmith thanks for adding this feature. I am wondering that is the cache keep updated after non-aligned saving?

william-silversmith:
Yes! Check out txrx.upload_image. In shade_and_upload, it calls upload_chunks which respects caching.

Issue #18, opened 2017-09-13
Read / Write compressed_segmentation


william-silversmith:
Reasons to support this: decrease storage amounts and remove decode step for neuroglancer viewing

PR #95, opened 2018-04-14
Python Compressed Segmentation Codec

william-silversmith:
compressed_segmentation is a format invented by JBMS for use with neuroglancer. It takes labels e.g. uint64s and renumbers the voxels inside to smaller numbers so that storage takes less space while affording random access. In tests, it typically comes out smaller than merely gzipped segmentation.

https://github.com/google/neuroglancer/tree/master/src/neuroglancer/sliceview/compressed_segmentation

This PR introduces to CloudVolume a pure python encoder/decoder by Yann Leprince at HBP.
https://github.com/HumanBrainProject/neuroglancer-scripts/blob/master/src/neuroglancer_scripts/_compressed_segmentation.py

There is a C++ version due to Stephen Plaza at Janelia that might be worth integrating with Cython here: https://github.com/janelia-flyem/compressedseg

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,6 +103,7 @@ ENV/
 # OS X 
 .DS_Store
 
+.pytest_cache
 test.py
 
 # Saved Images Cache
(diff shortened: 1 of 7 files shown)


william-silversmith:
Resolves #18

PR #98, opened 2018-04-15
feat: parallel uploads

### LINKED ISSUE #77: Parallel CloudVolume
Jingpeng and Sergiy have both requested that work on the parallel implementation be completed.

In Sergiy's case, it was because the gsutil implementation is hacky.

william-silversmith:
Currently paralellizes the aligned uploads for both ordinary and non-aligned uploads. 

Painting the non-aligned shell is currently handled by a single process.

The code does not yet handle low memory situations.

Resolves #77

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,9 +1,14 @@
 CHANGES
 =======
 
-0.15.2
+0.16.0
 ------
 
+* fix: updated setversion.py to not clobber
+* Python Compressed Segmentation Codec (#95)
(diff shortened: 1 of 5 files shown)


william-silversmith:
Dying on BOSS download due to server error. Try again tomorrow.

william-silversmith:
Things that remain to be done: 
- Parallelize the shell of non-aligned writes. This shouldn't be too hard with a little refactoring of the multiprocessing logic.
- Supply a way to provide shared-memory input to avoid the extra copy.

PR #102, opened 2018-04-19
feat: initial integration of Compresso algorithm

william-silversmith:
Compresso algorithm and code thanks to the Phister lab.
https://github.com/VCG/compresso/tree/master/paper

We might want to ask for help tuning the algorithm and implementation.


william-silversmith:
I've created https://github.com/seung-lab/compresso to facilitate this. This is so old we should do this in a new PR when the time is ripe.

PR #110, opened 2018-05-18
fix: make futures conditional on python 2

william-silversmith:


william-silversmith:
@wongwill86 reported this issue and confirmed that this fixes it.

PR #112, opened 2018-05-19
perf: 1x memory usage for parallel downloads

william-silversmith:
Previously, I was very conservative and wanted to ensure
that the usage of shared memory was completely invisible
to the end user of downloaded numpy arrays. However, this
required making a copy resulting in 2x memory usage.

I conducted some experiments showing that we can safely
avoid making this copy. Instead, we merely unlink the
shared memory. The memory region will remain accessible
so long as there is an open file descriptor to it. The FD
will be closed when the output VolumeCutout is garbage collected.

In all common use cases I have witnessed, this should result in
the transparent usage of shared memory without the concomitant
memory cost. In the case that a user fills an array with
thousands of VolumeCutouts, they have the option of making a
copy using np.copy and closing the file descriptors manually
or automatically.

### DIFF
--- a/cloudvolume/txrx.py
+++ b/cloudvolume/txrx.py
@@ -100,11 +100,8 @@ def multi_process_cutout(vol, requested_bbox, cloudpaths, parallel,
 
   mmap_handle, renderbuffer = shm.bbox2array(vol, requested_bbox, lock=fs_lock, location=shared_memory_location)
   if not output_to_shared_memory:
-    renderbuffer = np.copy(renderbuffer)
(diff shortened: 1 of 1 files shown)


william-silversmith:
parallel_download_mem
current

```
vol = CloudVolume('gs://neuroglancer/test_v0/segmentation', progress=True, parallel=2, cache=True)

for _ in range(300):
    img = vol[:]
    img[:100] + 1
    vol.shared_memory_id = vol.generate_shared_memory_location()
```

PR #121, opened 2018-06-25
feat: compressed segmentation C extension

william-silversmith:


william-silversmith:
Current decompression progress.

william-silversmith:
Decompression working. Compression progress:
screen shot 2018-07-09 at 12 42 54 am

william-silversmith:
Looks like it works when the block size is 8x8x8 and the chunk size is 64x64x64. Essentially, the above problem seems to be with the padding operation.

PR #137, opened 2018-09-12
feat(meshservice): Remove duplicate vertices

nkemnitz:
Removes duplicate vertex positions and relabels the triangle indices accordingly.


william-silversmith:
Nico said that it would be good to keep the option for expert users. There are situations in lighting and topology that require duplicate points.

nkemnitz:
Addendum: If we could modify the `unique` function used in this PR to only touch vertices at the chunk boundaries, it would mitigate the problem of accidentally merging vertices that have been kept separate on purpose.

PR #138, opened 2018-09-23
feat(meshservice): Faster mesh processing + PLY support

nkemnitz:
* Huge speedup for `decode_mesh_buffer`
* Added fast-ish `mesh_to_ply` conversion


william-silversmith:
Looks like 3D Studio Maxx and Blender both support ply. I think the change of default is okay.

PR #142, opened 2018-10-12
Easy External Multiprocessing

william-silversmith:
The provenance class derived from jsonschema objects did not pickle, preventing multiprocessing from operating correctly. This change preserves much of the same interface and validation characteristics of the DataLayerProvenance class while pickling correctly.

### DIFF
--- a/cloudvolume/cloudvolume.py
+++ b/cloudvolume/cloudvolume.py
@@ -31,7 +31,7 @@ from .lib import (
 from .meshservice import PrecomputedMeshService
 from .provenance import DataLayerProvenance
 from .skeletonservice import PrecomputedSkeletonService
-from .storage import SimpleStorage, Storage
+from .storage import SimpleStorage, Storage, reset_connection_pools
 from . import txrx
 from .volumecutout import VolumeCutout
 from . import sharedmemory
@@ -156,7 +156,20 @@ class CloudVolume(object):
     except:
       raise Exception("MIP {} has not been generated.".format(self.mip))
 
+  def __setstate__(self, d):
+    """Called when unpickling which is integral to multiprocessing."""
+    self.__dict__ = d 
+
+    if 'cache' in d:
+      self.init_submodules(d['cache'].enabled)
+    else:
+      self.init_submodules(False)
+    
+    # otherwise the pickle might have references to old connections
+    reset_connection_pools() 
+
   def init_submodules(self, cache):
+    """cache = path or bool"""
     self.cache = CacheService(cache, weakref.proxy(self)) 
     self.mesh = PrecomputedMeshService(weakref.proxy(self))
     self.skeleton = PrecomputedSkeletonService(weakref.proxy(self)) 
--- a/cloudvolume/cacheservice.py
+++ b/cloudvolume/cacheservice.py
@@ -282,6 +282,9 @@ class CacheService(object):
 
     return { 'local': already_have, 'remote': download_paths }
     
-
+  def __repr__(self):
+    return "CacheService(enabled={}, compress={}, path='{}')".format(
+      self.enabled, self.compress, self.path
+    )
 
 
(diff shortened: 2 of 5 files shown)


wongwill86:
A unit tests should be super easy to add here to test if you can:
1. pickle
2. run parallel processing see https://github.com/wongwill86/chunkflow/blob/scratch/tests/test_cloudvolume_datasource.py#L138 for simple example

I know at one point Cloudvolume manually handles the connection reset somewhere but if CV is to be used in an parallel process external to CloudVolume will it still work?
For an older version of cloudvolume i had to do this:
https://github.com/wongwill86/chunkflow/blob/scratch/src/chunkflow/cloudvolume_datasource.py#L73
if not maybe that could be a separate PR.

william-silversmith:
Good calls.

william-silversmith:
It looks like `__setstate__` is called after unpickling if defined. That might be a good place to call `reset_connection_pools`.

william-silversmith:
Important note: we are not currently supporting Python2 for multiprocessing, it appears to be significantly more tricky.

william-silversmith:
Gonna merge this. Please let me know if you run into any more problems with multiprocessing @wongwill86 and we can release a new version.

cloudvolume/cloudvolume.py
```diff
@@ -156,7 +156,20 @@ def __init__(self, cloudpath, mip=0, bounded=True, autocrop=False, fill_missing=
...
+    """Called when unpickling which is integral to multiprocessing."""
+    self.__dict__ = d 
+
+    if 'cache' in d:
+      self.init_submodules(d['cache'].enabled)
+    else:
+      self.init_submodules(False)
+    
+    # otherwise the pickle might have references to old connections
+    reset_connection_pools()
```
wongwill86: Is calling this global modifier is the right thing to do whenever we unpickle a single cloudvolume instance.
In the case of using a processpool we would be resetting the connection every time. Is that necessary?

Also, I don't know how `reset_connection_pools` work internally for s3 and gcs but is it multithread safe? i.e. when spawning processes that spawns threads (i think it's possible but not 100% sure)?

william-silversmith: That's a good point that unpickling != process spawning. It's not necessarily appropriate for all cases of unpickling. The code for resetting the spawning pools does this:

```
  S3_POOL = keydefaultdict(lambda service: keydefaultdict(lambda bucket_name: S3ConnectionPool(service, bucket_name)))
  GC_POOL = keydefaultdict(lambda bucket_name: GCloudBucketPool(bucket_name))
```

So it'll keep working no matter what, though it may lead to the accumulation of connections in certain cases (which can then crash the process if it happens enough). It's intended to be used on process spawn as anything in the pool is then invalid and we can ignore the contents with a clean conscience.

william-silversmith: The trick is to identify when the unpickling is a result of process spawning. Setting the global array to a new pool will cause the connections in the old pool to hang around until `__del__` is called. I don't always trust that to happen.

wongwill86: Right, just wanted to make sure that everything inside `S3/GCSConnectionPool` and stuff it calls is also thread safe. 

Did you check out the link to my hacked CV in my comment? lock + a global dict for the pid itself. it's a bit heavy handed so it works so far. Haven't spent too much time to check for other corner cases or more elegant methods though.

william-silversmith:
Oh I did see that but didn't understand why you were doing it. I see I
didn't think that anyone would want to put CloudVolume inside a thread vs a
process, but it could happen. Here's what I think of that:

If the unpickling happens in the same process, reusing the connection pool
is fine (no reset). If there's a new process, all the objects in the pool
are copied but disconnected from os level connections and a clear operation
is needed.

So I don't think a lock is necessary (and clearing in threads can even be
detrimental), but checking for a new PID is an excellent idea.

PR #145, opened 2018-10-16
fix: only reset connection pool when unpicking on a new PID

william-silversmith:
See discussion on #142


wongwill86:
👍 looks great!

william-silversmith:
The python2 issue comes from using urllib3==1.24 I'll set the requirement to 1.23

william-silversmith:
fwiw, looks like 1.24 was released 2 hrs ago

william-silversmith:
See discussion here: https://github.com/urllib3/urllib3/issues/1456

Requests is only compatible up to urllib3 1.23. Let's stick with 1.23 for now.

PR #150, opened 2018-11-04
Cutout Viewer

william-silversmith:
This PR introduces two new functions:

```python
img = vol[:]
img.viewer()
```

```python
from cloudvolume import visualize

img = np.random.uniform(low=0, high=1, size=(512, 512, 100)).astype(np.float64)
visualize(img, segmentation=False)
```

CloudVolume writes Neuroglancer compatible volumes, but sometimes setting up a an HTTP server and/or using neuroglancer has too much friction to use regularly. This can happen when:

- You are inspecting a pipeline and want to look at the input or output data for a job.
- You are experimenting with 3D numpy arrays unconnected to a dataset.

These cases were and still are served by generating PNG files using the `cloudvolume.lib.save_images` function and the `VolumeCutout.save_images` method. However, they do not provide: 

- The ability to instantly reslice along x,y, & z.
- Segmentation is written using its RGB values which can be very muddy to look at. Mapping segmentation to bright colors usually creates issues where two adjacent segments look identical but are not. Neuroglancer provides instance color randomization to address this issue. It also provides a way to easily access the underlying values.
- It is somewhat painful to map a pixel location to a neuroglancer volume from a PNG. 

Therefore, this viewer provides:

- The ability to reslice along x, y, and z.
- Segmentation recoloring with color randomization (is there a way to use the 4 color theorem?)
- Support for uint8, uint16, uint32, float32, and float64 data.
- Automatic computation of the neuroglancer translations of the coordinate you are hovering over.

We might consider providing automatic meshing.

The viewer works by launching a simple HTTP server from python that allows connections from localhost. It serves a small viewing application as HTML and Javascript which then requests data parameters and data which are then rendered in the application. 

The code in this PR is based on: https://github.com/seung-lab/data-cube-x


william-silversmith:
Screenshot of prototype image viewer:

image

Segmentation viewer:

image

william-silversmith:
More recent pics:

imgview
hyperview
segview

PR #151, opened 2018-11-04
Specifying mip level by voxel resolution

nicholasturner1:
* Adds an argument to the cloud-volume initialization so that users can explicitly specify the desired voxel resolution instead of needing to coordinate with the info file. This performs a quick conversion to mip index within the `__init__` function, so the change should be otherwise invisible. 

All tests still pass aside from `test_boss_download` which I probably just haven't configured on my workstation. If there's an easy way to do this I can re-run the test, but it seems unrelated.

### DIFF
--- a/cloudvolume/cloudvolume.py
+++ b/cloudvolume/cloudvolume.py
@@ -462,13 +462,20 @@ class CloudVolume(object):
   def mip(self, mip):
     mip = list(mip) if isinstance(mip, collections.Iterable) else int(mip)
     try:
-      if isinstance(mip, list):
+      if isinstance(mip, list):  # mip specified by voxel resolution
         self._mip = next((i for (i,s) in enumerate(self.scales)
                           if s["resolution"] == mip))
-      else:
+      else:  # mip specified by index into downsampling hierarchy
         self._mip = self.available_mips[mip]
+
     except:
-      raise Exception("MIP {} has not been generated.".format(mip))
+      if isinstance(mip, list):
+        available = self.available_resolutions
+      else:
+        available = self.available_mips
+      available_str = ", ".join(map(str, available)) 
+      msg = "MIP {} not found. Available: {}".format(mip, available_str)
+      raise IndexError(msg)
 
   @property
   def scales(self):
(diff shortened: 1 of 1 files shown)


cloudvolume/cloudvolume.py
```diff
@@ -105,7 +105,7 @@ class CloudVolume(object):
     non_aligned_writes: (bool) Enable non-aligned writes. Not multiprocessing safe without careful design.
       When not enabled, a ValueError is thrown for non-aligned writes.
   """
-  def __init__(self, cloudpath, mip=0, bounded=True, autocrop=False, fill_missing=False, 
+  def __init__(self, cloudpath, mip=0, resolution=None, bounded=True, autocrop=False, fill_missing=False,
```
nkemnitz: You should put the new `kwarg` at the very end, to ensure backward compatibility. Otherwise existing code, where the _keyword_ arguments are passed as _positional_ arguments, will start to behave strangely.

nkemnitz: Maybe even better, just handle both code paths with the single `mip` parameter? They describe the same thing, so it wouldn't make sense to use both at the same time; and their input types don't overlap.

william-silversmith:
This is awesome Nick! Thanks for contributing!

What would you think of having different behavior for mip based on the input type? int is previous behavior and an iterable type would trigger the resolution. I always hesitate to add new flags, byt let me know which way you'd find less confusing. After all, what happens if someone specifies resolution and mip?

nicholasturner1:
I wrote it so that resolution would override mip in case both are used. One seems more likely to know what they want if they specify the resolution than an index based on an info file they might not have created. 

Resolving by type seems fine. It does seem to be tempting confusion, but I can accept it if others find it more palatable.

cloudvolume/cloudvolume.py
nicholasturner1: Responded to these in the main thread - will do. Thanks for the feedback

william-silversmith:
We might want to make a setter for the `mip` property that will reflect this new logic. Otherwise people that do `vol.mip = [8,8,40]` will be sorely disappointed.

william-silversmith:
We should also throw a clear error if the matching mip level isn't found.

william-silversmith:
There should probably be a clearer error message if the resolution isn't found. Other than that, this looks good.

william-silversmith:
Awesome!

PR #158, opened 2018-11-28
vol.mesh.save() refactor path code

pgunn:
This changes the API for vol.mesh.save(), replacing the old filename parameter with a filepath parameter that's a full or relative path to the file you want to write (you should now also provide an extension). The old behaviour if you don't provide this parameter is retained.


william-silversmith:
Works for me.

william-silversmith:
Travis failures are due to credential issues, not related to this code.

PR #165, opened 2018-12-13
feat(storage): Batch requests for GCS files_exist()

nkemnitz:
Bundles blob exists queries into Batch requests (only GCS).


william-silversmith:
Nico claims a 3x speedup for GCS, but we both suspect that there is some bottleneck as it should be much faster than this.

PR #180, opened 2019-01-14
remove urllib3 requirement

jingpengw:
why do we need this specific version? urllib3 was not imported explicitly anywhere in this package.
I got a version conflict with JHU intern while building my docker image.


jingpengw:
the latest intern have higher version of urllib3

jingpengw:
I have made a few more changes to solve my version conflicts. this might not be usable for a general case.

PR #182, opened 2019-01-15
adding a chunk based duplicate vertex removal function

fcollman:
I wanted to start a dialog on this PR, its not functioning perfectly yet for reasons I'm not clear on, but it's designed to give the remove_duplicate_vertex functionality chunk boundary specificity, in order to avoid merging self touches but removing the cross chunk boundary artifacts in the mesh graph.

### DIFF
--- a/cloudvolume/meshservice.py
+++ b/cloudvolume/meshservice.py
@@ -23,6 +23,37 @@ def filename_to_segid(filename):
   segid, = matches.groups()
   return int(segid)
 
+def remove_duplicate_vertices_cross_chunks(verts, faces, chunk_size):
+    # find all vertices that are exactly on chunk_size boundaries
+    is_chunk_aligned = np.any(np.mod(verts, chunk_size) == 0, axis=1)
+    # # uniq_vertices, uniq_faces, vert_face_counts = np.unique(vertices[faces],
+    #                                                         return_inverse=True,
+    #                                                         return_counts=True,
+    #                                                         axis=0)
+    # find all vertices that have exactly 2 duplicates
+    unique_vertices, unique_inverse, counts = np.unique(verts,
+                                                        return_inverse=True,
+                                                        return_counts=True,
+                                                        axis=0)
+    only_double = np.where(counts == 2)[0]
+    is_doubled = np.isin(unique_inverse, only_double)
+    # this stores whether each vertex should be merged or not
+    do_merge = np.array(is_doubled & is_chunk_aligned)
+
+    # setup an artificial 4th coordinate for vertex positions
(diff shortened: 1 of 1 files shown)


fcollman:
Here's an example mesh that without remove_duplicate vertices is fragmented... (colored here by connected component)
screen shot 2019-01-14 at 7 53 10 pm
but the old way of merging vertices results in this self touch, visualized here by coloring the mesh with respect to distance along the mesh graph
screen shot 2019-01-14 at 8 13 01 pm
the new code however successfully merges, but not at this self touch
screen shot 2019-01-14 at 8 02 58 pm
I'm still getting some mistakes in the nucleus that I don't understand... perhaps i have the chunk size wrong... (colored again here by connected component)
screen shot 2019-01-14 at 7 59 54 pm

fcollman:
after some discussion and reflection with casey and sven, we realized that this case is because the edges of these nuclei represent very large sets of self touches, where the interior of the soma mesh and the nucleus mesh are perfectly touching all around the chunk boundaries.  This is a weird pathological case of a self touch that this code was designed not to merge, so we think it's acting appropriate for its design.

cloudvolume/meshservice.py
```diff
@@ -110,8 +142,13 @@ def produce_output(mdata):
       ])
 
       if remove_duplicate_vertices:
-        vertices, faces = np.unique(vertices[faces], return_inverse=True, axis=0)
-        faces = faces.astype(np.uint32)
+        if chunk_size:
+          vertices, faces = remove_duplicate_vertices_cross_chunks(vertices,
```
william-silversmith: What happens if you specify `remove_duplicate_vertices` but not `chunk_size`? Maybe remove_duplicate_vertices could be a boolean (in which case it removes all duplicate vertices) or a tuple (in which case it removes all fragment overlaps).

william-silversmith: Huh, I must have read the indentation wrong.

william-silversmith:
This will work most of the time, but it'll fail if the self-touch is on the boundary. Still, that's an improvement over never handling them at all.

fcollman:
yes agreed it will fail for self touches on the boundary, but I'm not sure what to do there, and in most cases (other than the nucleus case) the two compartments of a self touch will have duplicated vertices that will be removed and so the mesh will be connected.  There will be a virtual hole of 0 area right where the self touch and the mesh boundary intersect, but i'm not sure that's of real consequence.

I was quite proud of figuring out that 4th coordinate trick... which i also wouldn't have even considered if the clever unique trick to reindex the faces wasn't already there

Issue #170, opened 2018-12-17
feature: configurable s3-compliant protocols


RussTorres:
Currently support for s3-compliant apis is defined by adding options to the connectionpools module.  It is straightforward to modify this module to allow various s3-compliant storage systems (such as "matrix" and [wasabi](https://github.com/RussTorres/cloud-volume/commit/2c69d0f48f3547563ffd4d6a3bdb737f6563834c)), but it seems tedious to maintain personal forks for different endpoints -- at the Allen Institute we have either used or proposed using s3/cloudvolume with CEPH, Isilon, igneous, and wasabi.  An easier configuration which can use custom endpoint and e.g. boto profile information without modifying the codebase would be helpful.

william-silversmith:
This is a good idea Russel! There's a few ways we could implement it. One way might be to try detecting boto or aws configuration files, another might be to pass arguments to CloudVolume's constructor though it's a bit awkward since they would be arguments that apply for only one protocol. Have you used other means (i.e. not CloudVolume) of accessing these services before? How did you do it?

RussTorres:
In a few scripts I've had to target aws, wasabi, and igneous interchangeably.  I've done this using boto3 and the following `~/.aws/config`
```
[default]
region = us-west-2
[profile wasabi]
region = us-east-1
[profile igneous]
region = iggy-1
```
with a `~/.aws/credentials` like this:
```
[default]
aws_access_key_id = MYAWSKEY
aws_secret_access_key = MYAWSSECRET
[wasabi]
aws_access_key_id = MYWASABIKEY
aws_secret_access_key = MYWASABISECRET
[igneous]
aws_access_key_id = MYIGNEOUSKEY
aws_secret_access_key = MYIGNEOUSSECRET
```

The scripts in question have been built using argschema and a simple argschema parser built to attach to a designated s3 resource. -- A lot of this is the argschema/marshmallow boilerplate, but the `S3ResourceModule.__init__()` shows how to use the session and resource schemas defined above it.

```python
import os
try:
    from urllib.parse import urlparse
except ImportError:
    from urlparse import urlparse

import argschema
import boto3

from botocore.utils import fix_s3_host

example_input = {
    "resource": {
        "endpoint_url": "https://s3.wasabisys.com",
        "session": {
            "profile_name": "wasabi"
        }
    },
    "output_path": "s3://mybucket/",
    "input_fn": "example_txtfile.txt"
}

class S3Session(argschema.schemas.DefaultSchema):
[...]

PR #185, opened 2019-01-22
[Feature] add a little bit comments; add support for environment variables

jingpengw:
auto load the environment variable if there is no secret json files found.

### DIFF
--- a/cloudvolume/secrets.py
+++ b/cloudvolume/secrets.py
@@ -113,6 +113,14 @@ def aws_credentials(bucket = '', service = 'aws'):
       with open(aws_credentials_path, 'r') as f:
         aws_credentials = json.loads(f.read())
       break
+  
+  if not aws_credentials:
+    # did not find any secret json file, will try to find it in environment variables
+    if 'AWS_ACCESS_KEY_ID' in os.environ and 'AWS_SECRET_ACCESS_KEY' in os.environ:
+      aws_credentials = {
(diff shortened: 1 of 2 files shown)


cloudvolume/secrets.py
```diff
@@ -113,6 +113,14 @@ def aws_credentials(bucket = '', service = 'aws'):
       with open(aws_credentials_path, 'r') as f:
         aws_credentials = json.loads(f.read())
       break
+  
+  if not aws_credentials:
```
william-silversmith: I don't think I understand the use case here. When would an environment variable be present but not the file? I do think we should support CloudVolume accepting credentials in the constructor including the s3 endpoint per #170 Does this introduce any security issues that were not present with the files?

I guess AWS thinks it's ok: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html

jingpengw: environment variable could be transferred by `docker run -e`, and also defined in the kubernetes yml file. 
also boto3 and awscli will load environment variable too, so we only need to define environment variable and all the aws tools will work.

william-silversmith:
I think since it's a new feature its less likely to break things. We should
probably study s3 and Google so see how they handle it. People will
probably form their expectations from how those tools behave.

jingpengw:
at least, they will check the ~/.boto/credential file.

Jingpeng Wu
Postdoc in Princeton Neuroscience Institute, Princeton University, NJ
website: jingpengw.github.io

PR #188, opened 2019-01-23
Add from_numpy Function to Automatically Construct a Precomputed Volume

jingpengw:
- add from_numpy function to construct a volume; 
- remove unused imported packages and redefined functions

### DIFF
--- a/cloudvolume/cloudvolume.py
+++ b/cloudvolume/cloudvolume.py
@@ -245,7 +245,7 @@ class CloudVolume(object):
     resolution, voxel_offset, volume_size, 
     mesh=None, skeletons=None, chunk_size=(64,64,64),
     compressed_segmentation_block_size=(8,8,8),
-    mip_num=1 
+    mip_num=1, each_factor=Vec(2,2,1) 
   ):
     """
     Used for creating new neuroglancer info files.
@@ -266,30 +266,32 @@ class CloudVolume(object):
       compressed_segmentation_block_size: (x,y,z) dimensions of each compressed sub-block
         (only used when encoding is 'compressed_segmentation')
       mip_num: (int), the number of mip levels.
+      each_factor: (Vec), the downsampling factor
 
     Returns: dict representing a single mip level that's JSON encodable
     """
+    if not isinstance(each_factor, Vec):
+      each_factor = Vec(*each_factor)
+
     info = {
       "num_channels": int(num_channels),
       "type": layer_type,
       "data_type": data_type,
-    }
-    
-    # add mip levels
-    info['scales'] = []
-    for mip in range(mip_num):
-      resolution = (resolution[0]*2**mip, resolution[1]*2**mip, resolution[2])
-      voxel_offset = (voxel_offset[0]//2**mip, voxel_offset[1]//2**mip, voxel_offset[2])
-      volume_size = (volume_size[0]//2**mip, volume_size[1]//2**mip, volume_size[2])
-      scale = {
+      "scales": [{
         "encoding": encoding,
         "chunk_sizes": [chunk_size],
         "key": "_".join(map(str, resolution)),
         "resolution": list(map(int, resolution)),
         "voxel_offset": list(map(int, voxel_offset)),
         "size": list(map(int, volume_size)),
-      }
-      info['scales'].append(scale)
+      }],
+    }
+   
+    # add mip levels
+    factor = each_factor.clone()
+    for _ in range(1, mip_num):
+      cls.add_scale(factor, info=info)
+      factor *= each_factor
 
     if encoding == 'compressed_segmentation':
       info['scales'][0]['compressed_segmentation_block_size'] = list(map(int, compressed_segmentation_block_size))
--- a/cloudvolume/cloudvolume.py
+++ b/cloudvolume/cloudvolume.py
@@ -166,8 +166,6 @@ class CloudVolume(object):
     mip_num: (int) number of mip levels in the info file
     """
     path = lib.extract_path(vol_path)
-    if path.protocol == 'file':
-      mkdir(vol_path)
 
     if not layer_type:
       if arr.dtype in (np.uint8, np.float32, np.float16, np.float64):
--- a/cloudvolume/cloudvolume.py
+++ b/cloudvolume/cloudvolume.py
@@ -264,7 +264,7 @@ class CloudVolume(object):
       compressed_segmentation_block_size: (x,y,z) dimensions of each compressed sub-block
         (only used when encoding is 'compressed_segmentation')
       mip_num: (int), the number of mip levels.
-      each_factor: (Vec), the downsampling factor
+      factor: (Vec), the downsampling factor for each mip level
 
     Returns: dict representing a single mip level that's JSON encodable
     """
--- a/cloudvolume/cloudvolume.py
+++ b/cloudvolume/cloudvolume.py
@@ -163,7 +163,8 @@ class CloudVolume(object):
   def from_numpy(cls, arr, vol_path='file:///tmp/image/'+generate_random_string(),
                   resolution=(4,4,40), voxel_offset=(0,0,0), 
                   chunk_size=(128,128,64), layer_type=None):
-    if re.match('^file://', vol_path):
+    path = lib.extract_path(vol_path)
+    if path.protocol == 'file':
       mkdir(vol_path)
 
     if not layer_type:
--- a/cloudvolume/cloudvolume.py
+++ b/cloudvolume/cloudvolume.py
@@ -5,6 +5,7 @@ import collections
 import json
 import json5
 import os
+import re
 import sys
 import uuid
 import weakref
@@ -14,6 +15,7 @@ import numpy as np
 from tqdm import tqdm
 from six import string_types
 import multiprocessing as mp
+from time import strftime
 
 from intern.remote.boss import BossRemote
 from intern.resource.boss.resource import ChannelResource, ExperimentResource, CoordinateFrameResource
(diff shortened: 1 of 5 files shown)


william-silversmith:
This is a great idea.

jingpengw:
@william-silversmith can we merge?
Note that there are three `as_type` functions in Bbox class, and two of them were redefined. I deleted the first two and leave the last one. I am not 100% sure that this is the correct one. Please double check.

cloudvolume/cloudvolume.py
```diff
@@ -160,6 +157,27 @@ def __init__(self, cloudpath, mip=0, bounded=True, autocrop=False, fill_missing=
 
     self.pid = os.getpid()
 
+  @classmethod
+  def from_numpy(cls, arr, vol_path='file://tmp/'+generate_random_string(),
+                  resolution=(4,4,40), voxel_offset=(0,0,0), 
+                  chunk_size=(64,64,64)):
+    mkdir(vol_path)
+    if arr.dtype == np.uint8 or arr.dtype == np.float32:
```
william-silversmith: Unfortunately, these assumptions aren't necessarily true. I think they could be good defaults, but it's important to include a layer_type option in the arguments. For example, I've seen a sparse manually reconstructed dataset with a uint8 segmentation.

cloudvolume/cloudvolume.py
```diff
@@ -160,6 +157,27 @@ def __init__(self, cloudpath, mip=0, bounded=True, autocrop=False, fill_missing=
 
     self.pid = os.getpid()
 
+  @classmethod
+  def from_numpy(cls, arr, vol_path='file://tmp/'+generate_random_string(),
+                  resolution=(4,4,40), voxel_offset=(0,0,0), 
+                  chunk_size=(64,64,64)):
+    mkdir(vol_path)
```
william-silversmith: This only applies to the `file://` protocol.

cloudvolume/cloudvolume.py
```diff
@@ -160,6 +157,27 @@ def __init__(self, cloudpath, mip=0, bounded=True, autocrop=False, fill_missing=
...
+    if arr.dtype == np.uint8 or arr.dtype == np.float32:
+      layer_type = 'image'
+    elif arr.dtype == np.uint32 or arr.dtype == np.uint64:
+      layer_type = 'segmentation'
+
+    # cloud volume use fortran order, so we need to transpose 
+    # arr = np.transpose(arr)
+    # import pdb; pdb.set_trace()
```
william-silversmith: Remove debugging code.

cloudvolume/cloudvolume.py
william-silversmith: It's also important to consider uint16 and float64 for example. Integer volumes also exist.

cloudvolume/cloudvolume.py
```diff
@@ -160,6 +157,27 @@ def __init__(self, cloudpath, mip=0, bounded=True, autocrop=False, fill_missing=
...
+    info = cls.create_new_info(1, layer_type, arr.dtype.name, 'raw', resolution, voxel_offset, arr.shape)
```
william-silversmith: `num_channels=1` (I couldn't remember what it meant.)

cloudvolume/cloudvolume.py
```diff
@@ -160,6 +157,27 @@ def __init__(self, cloudpath, mip=0, bounded=True, autocrop=False, fill_missing=
 
     self.pid = os.getpid()
 
+  @classmethod
+  def from_numpy(cls, arr, vol_path='file://tmp/'+generate_random_string(),
+                  resolution=(4,4,40), voxel_offset=(0,0,0), 
+                  chunk_size=(64,64,64)):
```
william-silversmith: Maybe the chunk size should be (128,128,64) since (64,64,64) has typically been terrible for transfer rates.

cloudvolume/cloudvolume.py
```diff
@@ -160,6 +157,27 @@ def __init__(self, cloudpath, mip=0, bounded=True, autocrop=False, fill_missing=
...
+    if arr.dtype == np.uint8 or arr.dtype == np.float32:
+      layer_type = 'image'
+    elif arr.dtype == np.uint32 or arr.dtype == np.uint64:
+      layer_type = 'segmentation'
+
+    # cloud volume use fortran order, so we need to transpose 
+    # arr = np.transpose(arr)
+    # import pdb; pdb.set_trace()
+    info = cls.create_new_info(1, layer_type, arr.dtype.name, 'raw', resolution, voxel_offset, arr.shape)
+    vol = CloudVolume(vol_path, info=info, bounded=True, autocrop=True)
```
william-silversmith: I'm not sure what `autocrop` is buying us here since the size of the volume is exactly equal to the new volume. Maybe we should turn it off?

cloudvolume/cloudvolume.py
```diff
@@ -160,6 +157,27 @@ def __init__(self, cloudpath, mip=0, bounded=True, autocrop=False, fill_missing=
...
+    elif arr.dtype == np.uint32 or arr.dtype == np.uint64:
+      layer_type = 'segmentation'
+
+    # cloud volume use fortran order, so we need to transpose 
+    # arr = np.transpose(arr)
+    # import pdb; pdb.set_trace()
+    info = cls.create_new_info(1, layer_type, arr.dtype.name, 'raw', resolution, voxel_offset, arr.shape)
+    vol = CloudVolume(vol_path, info=info, bounded=True, autocrop=True) 
+    # save the info file
+    vol.commit_info()
```
william-silversmith: We should set the provenance file automatically if at all possible. We could set the processing field to say it was created from this method for example.

william-silversmith:
As for `astype`... I'm not sure how that happened. As far as I can tell, none of them are right. Here's the right one...

```python
  def astype(self, typ):
    tmp = self.clone()
    tmp.minpt = tmp.minpt.astype(typ)
    tmp.maxpt = tmp.maxpt.astype(typ)
    tmp._dtype = tmp.minpt.dtype 
    return tmp
```

cloudvolume/cloudvolume.py
jingpengw: em, if we want to support multiple channel, I should make it more general.

cloudvolume/cloudvolume.py
jingpengw: I see, I have to use `/tmp/image/dfds` rather than `/tmp/dfds`

cloudvolume/cloudvolume.py
jingpengw: I have make it a little bit more smarter.

cloudvolume/cloudvolume.py
jingpengw: agree.

cloudvolume/cloudvolume.py
jingpengw: modified as suggested

cloudvolume/cloudvolume.py
jingpengw: added

jingpengw:
@william-silversmith tks for the careful review, I have resolved all the concerns, please take a look.

cloudvolume/lib.py
```diff
@@ -817,7 +812,7 @@ def normalize_float(img):
     elif index == 2:
       img = image[:, :, level, channel ]
     else:
-      raise NotImplemented
+      raise NotImplementedError
```
jingpengw: @william-silversmith please double check whether this is appropriate:
https://docs.python.org/3/library/constants.html#NotImplemented

william-silversmith: Thanks for that link. I didn't know that NotImplemented was defined as a constant. According to https://docs.python.org/3/library/exceptions.html

```This exception is derived from RuntimeError. In user defined base classes, abstract methods should raise this exception when they require derived classes to override the method, or while the class is being developed to indicate that the real implementation still needs to be added.```

This would imply that your correction from NotImplemented to NotImplementedError is more correct than what I wrote, but that we should probably use some other kind of error like IndexError or a custom error.

File: cloudvolume/lib.py [hunk 9]
```diff
@@ -154,7 +154,7 @@ def extract_path(cloudpath):
 def toabs(path):
   home = os.path.join(os.environ['HOME'], '')
   # remove file protocol
-  path = re.sub('^file://?', '', path)
+  # path = re.sub('^file://?', '', path)
```
william-silversmith: Is this intended to be commented out?

cloudvolume/cloudvolume.py
```diff
@@ -14,6 +15,7 @@
 from tqdm import tqdm
 from six import string_types
 import multiprocessing as mp
+from time import strftime
```
william-silversmith: Is this actually being used?

cloudvolume/cloudvolume.py
```diff
@@ -158,22 +160,36 @@ def __init__(self, cloudpath, mip=0, bounded=True, autocrop=False, fill_missing=
...
-    elif arr.dtype == np.uint32 or arr.dtype == np.uint64:
-      layer_type = 'segmentation'
-
-    # cloud volume use fortran order, so we need to transpose 
-    # arr = np.transpose(arr)
-    # import pdb; pdb.set_trace()
-    info = cls.create_new_info(1, layer_type, arr.dtype.name, 'raw', resolution, voxel_offset, arr.shape)
-    vol = CloudVolume(vol_path, info=info, bounded=True, autocrop=True) 
+                  chunk_size=(128,128,64), layer_type=None):
+    if re.match('^file://', vol_path):
```
william-silversmith: It might be more future proof to do something like:

```python
path = lib.extract_path(vol_path)
if path.protocol == 'file':
    mkdir(vol_path)
```

cloudvolume/lib.py
jingpengw: yes, originally there is no such removal. I should delete this rather than commented out.

cloudvolume/cloudvolume.py
jingpengw: yes, in the provenance generation:
https://github.com/seung-lab/cloud-volume/pull/188/files/cb58e847324315c7dbd5e84b44395df60181ff3c..419da54ecde51217d61f75675fd20c0031b88d21#diff-7095eb78d90e99a1cebf3f6e0c41a094R190

cloudvolume/cloudvolume.py
jingpengw: modified as suggested.

cloudvolume/cloudvolume.py
```diff
@@ -261,22 +265,31 @@ def create_new_info(cls,
...
       "num_channels": int(num_channels),
       "type": layer_type,
       "data_type": data_type,
-      "scales": [{
+    }
+    
+    # add mip levels
+    info['scales'] = []
+    for mip in range(mip_num):
+      resolution = (resolution[0]*2**mip, resolution[1]*2**mip, resolution[2])
```
william-silversmith: This is good for most use cases, but only works if your mip levels are 2x2x1 downsampled. The more authoritative logic is located in Igneous in downsample_scales. We could integrate there here maybe?

william-silversmith: On second thought this is fine. In the future, we should add a "factor" parameter with it set to (2,2,1) by default.

cloudvolume/cloudvolume.py
```diff
@@ -226,22 +265,31 @@ def create_new_info(cls,
...
+      voxel_offset = (voxel_offset[0]//2**mip, voxel_offset[1]//2**mip, voxel_offset[2])
+      volume_size = (volume_size[0]//2**mip, volume_size[1]//2**mip, volume_size[2])
```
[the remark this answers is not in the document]
william-silversmith: Yes, check out `def add_scale` for the correct logic. voxel_offset uses `np.floor` and size uses `np.ceil`.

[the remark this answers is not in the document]
jingpengw: just fixed this. using existing add_scale function now.

cloudvolume/cloudvolume.py
jingpengw: good catch. using existing add_scale function now.

jingpengw:
@william-silversmith could you take another look?

william-silversmith:
One last thing 😅 can we change "each_factor" to "factor"? Shorter and it
keeps in concert with similar functions. Thank you for doing this!

jingpengw:
@william-silversmith tks for the careful review! this will make the package easier to use and maintain.

cloudvolume/cloudvolume.py
```diff
@@ -243,7 +243,7 @@ def create_new_info(cls,
     resolution, voxel_offset, volume_size, 
     mesh=None, skeletons=None, chunk_size=(64,64,64),
     compressed_segmentation_block_size=(8,8,8),
-    mip_num=1, each_factor=Vec(2,2,1) 
+    mip_num=1, factor=Vec(2,2,1) 
   ):
     """
     Used for creating new neuroglancer info files.
```
william-silversmith: The docstring still has `each_factor` in it.

jingpengw:
woo hoo!

Issue #68, opened 2018-02-12
Provide Fortran Order Flag


william-silversmith:
We can decrease memory requirements a lot by not performing the transposes necessary to convert from fortran to C order. Ran Lu for instance wants to work with Fortran order anyway so it's just wasted work. Provide a flag, e.g. order='C' or order='F' like numpy

jingpengw:
why do we close this? Ran do not need it anymore?
I am using C order in chunkflow, and was always doing transposing before and after inference. The transpose of affinitymap increased the memory overhead a lot. I would like to make the order flag work. 
I can work on this and create a PR.

william-silversmith:
This was closed because all precomputed volumes are in F order. We've been
trying to not upload unviewable volumes.

Could you describe your use case?

jingpengw:
we can use C order if you just rename the xyz to zyx?
My packages are all based on C order numpy array, so I have to transpose before or after using cloudvolume.

jingpengw:
the volume should be viewable if we use C order with `c z y x` indexing. The memory layout will be exactly the same with neuroglancer precomputed and the indexing style will be consistent with numpy.

jingpengw:
I have PR to work on this:
https://github.com/seung-lab/cloud-volume/pull/192

william-silversmith:
I wrote fastremap, which offers in-place transposition to avoid the memory overhead for 2D, 3D, and 4D rectangular arrays.

https://github.com/seung-lab/fastremap

jingpengw:
maybe we can try to use `fastremap` to implement this flag.

william-silversmith:
I was initially worried that doing that forces users to use in-place transposition, which can take longer than out of place, but a user could also just not set the flag and do out-of-place transposition outside of CloudVolume. It does also create an additional dependency, but supporting `graphene` will probably require fastremap anyway.

william-silversmith:
Actually, while it's clear what to do for downloads, uploading is more tricky. If we applied an IPT to the upload, it would change the input the user feeds to CloudVolume. We could apply it twice, but that would be slow, and there would be a gap where the data was changing which could be weird in a multithreaded  environment (e.g. for shared memory).

william-silversmith:
I think I'm going to close this and just mention fastremap in the documentation for CloudVolume.

william-silversmith:
Added some documentation. 

[...]

PR #192, opened 2019-01-25
[WIP] C order flag

jingpengw:
this PR is intended to add a `order` flag to work with C order. I was supposed to be backward compatible, so the default working mode is fortran order.

this is not urgent, so I will stop working on it. Just put it here in case anyone would like to continue working on it.

currently, there are two problems I encountered:
- the C order shared memory upload is not working; 
- the chunksize in info is not integer but float number;


william-silversmith:
Based on the reasoning in #68 I'm going to close this. We can reopen it if someone objects.

PR #194, opened 2019-01-31
experiment: CloudVolume with Green Threads

william-silversmith:
- Using gevent
- gevent requires monkeypatching stdlib to achieve async io.
- gevent can use libev or libuv under the hood. I imagine there might be some slightly different performance characteristics. On Linux, it defaults to libev as libuv binaries are not easily available. Getting libuv requires compiling from source.

### DIFF
--- a/cloudvolume/storage/storage.py
+++ b/cloudvolume/storage/storage.py
@@ -226,7 +226,12 @@ def schedule_green_jobs(
     progress=None, total=None
   ):
 
-  total = len(fns) if total is None else total
+  if total is None:
+    try:
+      total = len(fns)
+    except TypeError: # generators don't have len
+      pass
+
   pbar = tqdm(total=total, desc=progress, disable=(not progress))
   results = []
   
@@ -651,9 +656,5 @@ class ThreadedStorage(AbstractStorage, ThreadedQueue):
   def __exit__(self, exception_type, exception_value, traceback):
     ThreadedQueue.__exit__(self, exception_type, exception_value, traceback)
     self._interface.release_connection()
-
-# Define alias for Storage
-# Storage = ThreadedStorage
-Storage = GreenStorage
 
 
--- a/cloudvolume/storage/storage.py
+++ b/cloudvolume/storage/storage.py
@@ -1,0 +1,489 @@
+import six
+from six.moves import queue as Queue
+from collections import defaultdict
+import json
+import os.path
+import re
+from functools import partial
+
+from tqdm import tqdm
+
+from cloudvolume import compression
+from cloudvolume.exceptions import UnsupportedProtocolError
+from cloudvolume.lib import mkdir, extract_bucket_path, scatter
+from cloudvolume.threaded_queue import ThreadedQueue
+
+from .storage_interfaces import (
+  FileInterface, HttpInterface, 
+  S3Interface, GoogleCloudStorageInterface
+)
+
+import gevent
+import gevent.pool
+
+DEFAULT_THREADS = 20
+
+def get_interface_class(protocol):
+  if protocol == 'file':
+    return FileInterface
+  elif protocol == 'gs':
+    return GoogleCloudStorageInterface
+  elif protocol in ('s3', 'matrix'):
+    return S3Interface
+  elif protocol in ('http', 'https'):
+    return HttpInterface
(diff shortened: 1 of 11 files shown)


william-silversmith:
Gevent socket monkey patch breaks Python 2.7 multiprocessing https://stackoverflow.com/questions/8678307/gevent-monkeypatching-breaking-multiprocessing

jingpengw:
what's the difference with current python threading?

william-silversmith:
We discard preemptible kernel threads and use a userspace cooperative
threading model. This should reduce coordination overhead and reduce CPU
thrashing, especially on virtualized or NUMA architectures.

cloudvolume/__init__.py
```diff
@@ -1,3 +1,9 @@
+import sys
+
+if sys.version_info[0] >= 3:
+  import gevent.monkey
+  gevent.monkey.patch_all()
```
nkemnitz: Maybe add a warning about the monkey patching as it might affect other parts of their program and lead to weird issues.

test/test_cloudvolume.py
```diff
@@ -125,7 +125,7 @@ def test_ellipsis_read():
 def test_parallel_read():
   paths = [
     'gs://seunglab-test/test_v0/image',
-    's3://seunglab-test/test_v0/image',
+    # 's3://seunglab-test/test_v0/image',
```
nkemnitz: Keep test for s3

test/test_storage.py
```diff
@@ -11,13 +11,14 @@
 #TODO delete files created by tests
 def test_read_write():
   urls = [
-    "file:///tmp/removeme/read_write",
+    # "file:///tmp/removeme/read_write",
```
nkemnitz: Keep test for file interface

cloudvolume/storage/storage.py
```diff
@@ -0,0 +1,628 @@
...
+        self.put(partial(exist_thunk, block))
+    else:
+      exist_thunk(file_paths, self._interface)
+
+    desc = 'Existence Testing' if self.progress else None
+    self.wait(desc)
+
+    return results
+
+  def get_json(self, file_path):
```
nkemnitz: Looks like this already exists in `AbstractStorage`.

test/test_cloudvolume.py
```diff
@@ -11,7 +11,7 @@
 from functools import reduce
 
 from cloudvolume.exceptions import AlignmentError
-from cloudvolume import CloudVolume, chunks, Storage
+from cloudvolume import CloudVolume, chunks, Storage, view
```
nkemnitz: `view` is unused

william-silversmith:
Yet another way would be to have the cloud-volume PyPI package install two packages: cloudvolume and ecocloudvolume (perhaps that name needs some work...). They would be functionally identical packages, except that only ecocouldvolume triggers monkeypatching and green threads.

william-silversmith:
Did an experiment on a GCE n1-standard-16. I found that between parallel 1-8, green thread throughput scaled linearly from 29 to 60 MVx/sec as `y = 18.4x + 29`. From parallel 8-16, throughput scaled as `y = 4.25x + 176`.

Ordinary cloudvolume followed a superlinear curve. The least squares best fit quadratic regression was `
0.242252 x^2 + 5.45288 x + 2.47059`. 

Parallel | Threaded (MVx/sec) | Green (MVx/sec)
------ | ------ | ------
16 | 152 | 210
12 |  99 | 188
8 | 60 | 176
4 | 30 | 103
2 | 16 | 56
1 | 6 | 29

We found that task pinning (`settask -c 0 python perf.py`) causes threaded cloudvolume to match green cloudvolume on a single core task. Therefore, my interpretation of the above equation is that the linear term is roughly equal to the linear improvement seen in the green threads equation, but the effect is suppressed by the quadratic thread scheduling overhead term.

william-silversmith:
green_threads_img_comparison
green_threads_segmentation_comparison

william-silversmith:
I think the right way to do this is to a) make the user consciously use monkey patching (it has to be at the start of a program anyway) and b) probably use the coming remix / plugin system

william-silversmith:
This is handled in #230

PR #197, opened 2019-02-07
fix from numpy function with multiple mip levels

jingpengw:
the usage of add_scale function is not correct.
also make the test creating multiple mip levels of the info.

### DIFF
--- a/cloudvolume/cloudvolume.py
+++ b/cloudvolume/cloudvolume.py
@@ -161,9 +161,12 @@ class CloudVolume(object):
   @classmethod
   def from_numpy(cls, arr, vol_path='file:///tmp/image/'+generate_random_string(),
                   resolution=(4,4,40), voxel_offset=(0,0,0), 
-                  chunk_size=(128,128,64), layer_type=None, mip_num=1):
+                  chunk_size=(128,128,64), layer_type=None, max_mip=0):
     """
-    mip_num: (int) number of mip levels in the info file
+    max_mip: (int) the maximum mip level id in the info file. 
+    Note that currently the numpy array can only sit in mip 0,
+    the max_mip was only created in info file.
+    the numpy array itself was not downsampled. 
     """
     path = lib.extract_path(vol_path)
 
@@ -183,7 +186,7 @@ class CloudVolume(object):
       raise NotImplementedError
 
     info = cls.create_new_info(num_channels, layer_type, arr.dtype.name, 'raw', resolution, 
-                               voxel_offset, arr.shape[:3], chunk_size=chunk_size, mip_num=mip_num)
+                               voxel_offset, arr.shape[:3], chunk_size=chunk_size, max_mip=max_mip)
     vol = CloudVolume(vol_path, info=info, bounded=True, autocrop=False) 
     # save the info file
     vol.commit_info()
@@ -243,7 +246,7 @@ class CloudVolume(object):
     resolution, voxel_offset, volume_size, 
     mesh=None, skeletons=None, chunk_size=(64,64,64),
     compressed_segmentation_block_size=(8,8,8),
-    mip_num=1, factor=Vec(2,2,1) 
+    max_mip=0, factor=Vec(2,2,1) 
   ):
     """
     Used for creating new neuroglancer info files.
@@ -263,7 +266,7 @@ class CloudVolume(object):
       chunk_size: int (x,y,z), dimensions of each downloadable 3D image chunk in voxels
       compressed_segmentation_block_size: (x,y,z) dimensions of each compressed sub-block
         (only used when encoding is 'compressed_segmentation')
-      mip_num: (int), the number of mip levels.
+      max_mip: (int), the maximum mip level id.
       factor: (Vec), the downsampling factor for each mip level
 
     Returns: dict representing a single mip level that's JSON encodable
@@ -293,7 +296,7 @@ class CloudVolume(object):
     factor_in_mip = factor.clone()
  
     # add mip levels
-    for _ in range(1, mip_num):
+    for _ in range(max_mip):
       new_resolution = list(map(int, Vec(*fullres['resolution']) * factor_in_mip ))
       newscale = {
         u"encoding": encoding,
--- a/cloudvolume/cloudvolume.py
+++ b/cloudvolume/cloudvolume.py
@@ -284,12 +284,25 @@ class CloudVolume(object):
         "size": list(map(int, volume_size)),
       }],
     }
-   
-    # add mip levels
+    
+    def downscale(size, factor_in_mip, roundingfn):
+      smaller = Vec(*size, dtype=np.float32) / Vec(*factor_in_mip)
+      return list(map(int, roundingfn(smaller)))
+    
+    fullres = info['scales'][0]
     factor_in_mip = factor.clone()
+ 
+    # add mip levels
     for _ in range(1, mip_num):
-      cls.add_scale(cls, factor_in_mip, info=info)
       factor_in_mip *= factor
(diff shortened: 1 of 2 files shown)


jingpengw:
the bug is from python 2.7.
is there any plan to drop support of python2.7?

jingpengw:
@william-silversmith fixed the bug.

jingpengw:
I have some minor correction commits, it might be better to squash the commits and merge.
could you also release a new version of CloudVolume after merging this?

cloudvolume/cloudvolume.py
```diff
@@ -161,14 +161,17 @@ def __init__(self, cloudpath, mip=0, bounded=True, autocrop=False, fill_missing=
...
+    max_mip: (int) the maximum mip level id in the info file. 
+    Note that currently the numpy array can only sit in mip 0,
+    the max_mip was only created in info file.
+    the numpy array itself was not downsampled. 
     """
     path = lib.extract_path(vol_path)
 
     if not layer_type:
-      if arr.dtype in (np.uint8, np.float32, np.float16, np.float64):
+      if arr.dtype in (np.bool, np.uint8, np.float32, np.float16, np.float64):
```
william-silversmith: What about np.intX? They'll creep in eventually, might as well handle it now.

cloudvolume/cloudvolume.py
```diff
@@ -183,7 +186,7 @@ def from_numpy(cls, arr, vol_path='file:///tmp/image/'+generate_random_string(),
       raise NotImplementedError
 
     info = cls.create_new_info(num_channels, layer_type, arr.dtype.name, 'raw', resolution, 
-                               voxel_offset, arr.shape[:3], chunk_size=chunk_size, mip_num=mip_num)
+                               voxel_offset, arr.shape[:3], chunk_size=chunk_size, max_mip=max_mip)
     vol = CloudVolume(vol_path, info=info, bounded=True, autocrop=False)
```
william-silversmith: Autocrop is False by default as is bounded = True. Why overspecify?

cloudvolume/cloudvolume.py
```diff
@@ -161,14 +161,17 @@ def __init__(self, cloudpath, mip=0, bounded=True, autocrop=False, fill_missing=
   @classmethod
   def from_numpy(cls, arr, vol_path='file:///tmp/image/'+generate_random_string(),
                   resolution=(4,4,40), voxel_offset=(0,0,0), 
-                  chunk_size=(128,128,64), layer_type=None, mip_num=1):
+                  chunk_size=(128,128,64), layer_type=None, max_mip=0):
```
william-silversmith: We should be able to pass in starting mip and `compress` as well. For what it's worth, around the rest of the codebase, I usually refer to the `mip_num` / `max_mip` parameter as `num_mips`. Might be good to keep it that way for consistency.

cloudvolume/cloudvolume.py
```diff
@@ -263,7 +266,7 @@ def create_new_info(cls,
       chunk_size: int (x,y,z), dimensions of each downloadable 3D image chunk in voxels
       compressed_segmentation_block_size: (x,y,z) dimensions of each compressed sub-block
         (only used when encoding is 'compressed_segmentation')
-      mip_num: (int), the number of mip levels.
+      max_mip: (int), the maximum mip level id.
```
william-silversmith: I typically call this parameter `num_mips`. We should be careful about introducing new parameters then changing their names... That will break other people's code (eventually). It's probably okay here because it's so new.

cloudvolume/cloudvolume.py
```diff
@@ -284,11 +287,26 @@ def create_new_info(cls,
         "size": list(map(int, volume_size)),
       }],
     }
-   
-    # add mip levels
+    
+    def downscale(size, factor_in_mip, roundingfn):
```
william-silversmith: I don't like replicating this logic by copy paste. Would it make more sense to refactor this into a function and use it here and in add_scale?

cloudvolume/cloudvolume.py
william-silversmith: It would be good to be able to pass in `voxel_offset` too.

william-silversmith:
On Python2.7, within the lab, undergrads and first years often use whatever is installed on their system by default. Outside the lab, I'm not sure what people are using, and I wouldn't be surprised if Python 2.7 is common. There are a few places where supporting Python 2.7 is a significant impact.

- Transferring buffers from Cython to Python incurs a memory copy
- Python 2.7 plays poorly with parallel green threads
- We have to use the six library to provide inter-major-version compatibility.

However, so far this has not been substantial enough of an impact to make supporting 2.7 take more than a few more minutes of work here and there. I'm planning on dropping support for 2.7 a few months after EOL on January 1, 2020. Numpy will probably drop support then which will start forcing people to migrate. If we encounter a substantive issue, I'm willing to reconsider this position.

cloudvolume/cloudvolume.py
jingpengw: I am using np.issubdtype function to include them all!

cloudvolume/cloudvolume.py
jingpengw: deleted the autocrop option.

cloudvolume/cloudvolume.py
jingpengw: I noticed that we need to use `+1` if using num_mips, I hate things like this. It is easier to introduce off by 1 bugs.

cloudvolume/cloudvolume.py
jingpengw: I don't either. I should follow the principles of clean code. 
moved this function outside the class.

cloudvolume/cloudvolume.py
jingpengw: 1. there exists `voxel_offset` parameter
2. added an encoding parameter
3. have not changed the max_mip parameter since we need to use +1 if using num_mips.

william-silversmith: How did I miss the voxel_offset... :man_facepalming:

cloudvolume/cloudvolume.py
william-silversmith: This is a good idea. Thanks for explaining the reasoning.

cloudvolume/cloudvolume.py
william-silversmith: Why not add `compress` as well as encoding though? `encoding` is e.g. `fpzip` or `compressed_segmentation` while compress is a bitstream compressor e.g. `gzip` (or potentially something like `lz4` in the future).

jingpengw: added. 
But is compress not used anywhere?
I have a question here:
https://github.com/seung-lab/cloud-volume/issues/198

cloudvolume/cloudvolume.py
```diff
@@ -161,17 +166,18 @@ def __init__(self, cloudpath, mip=0, bounded=True, autocrop=False, fill_missing=
...
     """
-    path = lib.extract_path(vol_path)
-
     if not layer_type:
-      if arr.dtype in (np.bool, np.uint8, np.float32, np.float16, np.float64):
+      if arr.dtype in (np.uint32, np.uint64, np.uint16):
+        layer_type = 'segmentation'
+      elif np.issubdtype(arr.dtype, np.integer) or np.issubdtype(arr.dtype, np.float):
         layer_type = 'image'
       elif arr.dtype in (np.uint32, np.uint64, np.uint16):
```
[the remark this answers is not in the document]
jingpengw: good catch!
deleted.

PR #204, opened 2019-03-12
changing to an os independent character

fcollman:
the old version is broken for windows.


william-silversmith:
Thanks Forrest! This looks fine, but while I don't run Windows myself, my impression was `posix_ipc` would cause installation to fail on Windows. Are you using CV on windows successfully?

PR #208, opened 2019-03-22
GrapheneMeshService able to decode draco meshes [for graphene branch]

manuel-castro:
Adds DracoPy as a dependency. Will attempt to use DracoPy to decode a
fragment then fall back to the old decoder if that raises an error.

### DIFF
--- a/cloudvolume/meshservicegraphene.py
+++ b/cloudvolume/meshservicegraphene.py
@@ -15,6 +15,8 @@ from .lib import red, toiter
 from .storage import Storage
 from cloudvolume import meshservice
 
+import DracoPy
+
 SEGIDRE = re.compile(r'/(\d+):0.*?$')
 
 def remove_duplicate_vertices_cross_chunks(verts, faces, chunk_size):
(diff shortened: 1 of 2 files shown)


cloudvolume/meshservicegraphene.py
```diff
@@ -146,7 +169,11 @@ def get(self, seg_id, remove_duplicate_vertices=False):
         meshdata = []
         for frag in tqdm(fragments, disable=(not self.vol.progress),
                          desc="Decoding Mesh Buffer"):
-            mesh = decode_mesh_buffer(frag["content"])
+            try:
+                # EAFP
```
william-silversmith: What does EAFP mean?

requirements.txt
```diff
@@ -2,6 +2,7 @@ backports-abc==0.5
 boto3>=1.4.7
 chardet==3.0.4
 compressed-segmentation>=1.0.0
+DracoPy>=0.0.5
```
william-silversmith: This is great! However, people often complain when I add additional compilation steps to CloudVolume. Would you be interested in creating binaries for DracoPy for Linux? I had trouble compiling on MacOS the other day though I didn't dig very hard.

cloudvolume/meshservicegraphene.py
william-silversmith: Thinking about this a bit more, I think it would be cleaner to provide a `DracoPy.is_draco_file` method that returns a boolean, or throw a custom exception like DracoPy.FileTypeException as a ValueError could be generated internal to Draco and provides less certainty that taking the standard path is actually correct.

cloudvolume/meshservicegraphene.py
```diff
@@ -65,8 +67,29 @@ def decode_mesh_buffer(fragment):
...
+    assert len(vertices) % 3 == 0, "Draco mesh vertices not 3-D"
+    num_vertices = len(vertices) // 3
+
+    # For now, just return this dict until we figure out
+    # how exactly to deal with Draco's lossiness/duplicate vertices
+    return {
+        'num_vertices': num_vertices,
+        'vertices': vertices.reshape(num_vertices, 3),
+        'faces': faces,
+        'was_draco_encoded': True
```
william-silversmith: Maybe it would be better to have 'encoding': 'draco' | 'precomputed' in case we'd like to use more encodings.

manuel-castro: Agreed. I will fix this in the next commit

cloudvolume/meshservicegraphene.py
manuel-castro: Agreed. I will fix this in the next commit

requirements.txt
manuel-castro: So I looked into creating manylinux binaries today. I had trouble building draco on manylinux/CentOS 5. I think the gcc compiler on CentOS 5 might be too old. Could we hold off on this until manylinux2010 becomes standard? It will drop CentOS 5 support. It seems relatively close: https://github.com/pypa/manylinux/issues/179

william-silversmith: @sdorkenw this impacts you the most, what do you think? Do we need some kind of exception logic in case DracoPy can't compile so that you can still use Cloudvolume _Graphene Collectors' Edition_ without it?

PR #209, opened 2019-03-29
misc enhancements

shangmu:
aws_credentials: likely less mystic error from calling code if no credential is available.
google_credentials: potentially enabling retries in the future


william-silversmith:
Rebased on master to see if tests pass.

PR #212, opened 2019-04-11
feat: delete black uploads flag

william-silversmith:
It's often wasteful to upload completely zeroed (black) files. Additionally, certain storage systems choke on many extremely tiny files, e.g. when they apply erasure coding, which splits 300 bytes into even tinier pieces and then creates replicated metadata records for them.  

This PR creates a new constructor flag and CloudVolume instance attribute: `delete_black_uploads=bool`, which is by default False.

When enabled, this feature issues a DELETE http request instead of a PUT for chunks that would be all zeros. This covers two cases:

1) There was no file to begin with: Good, we won't upload a file or create a record.
2) There was a preexisting file: Good, it's deleted now. We would have overwritten it with zeros.

This update also contains two other additions of note:

1) Ability to upload scalars with CloudVolume (i.e. `vol[...] = 3`)
2) Refactoring away of create_volume_from_image and replace it with `CloudVolume.from_numpy`

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,5 +1,6 @@
 Forrest Collman <forrest.collman@gmail.com>
 Jingpeng Wu <jingpeng.wu@gmail.com>
+Manuel Castro <macastro@princeton.edu>
 Nicholas Turner <nturner.stanford@gmail.com>
 Nico Kemnitz <nkemnitz@princeton.edu>
 Pat Gunn <pgunn01@gmail.com>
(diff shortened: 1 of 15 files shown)


william-silversmith:
Actually, that last fix is not necessary. We're already testing for it. 

```
delete_object(**kwargs)
Removes the null version (if there is one) of an object and inserts a delete marker, which becomes the latest version of the object. If there isn't a null version, Amazon S3 does not remove any objects.
```

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.delete_object

PR #215, opened 2019-05-06
Add allow_cropped flag to CloudVolume.download_point

manuel-castro:
Download_point can raise a ValueError if the bbox is out of bounds. This PR adds a flag to allow a cropped bbox, and resets self.mip if a ValueError happens.

### DIFF
--- a/cloudvolume/cloudvolume.py
+++ b/cloudvolume/cloudvolume.py
@@ -8,6 +8,7 @@ import os
 import sys
 import uuid
 import weakref
+import traceback
 
 from six.moves import range
 import numpy as np
@@ -1032,7 +1033,7 @@ class CloudVolume(object):
 
     return self._boss_cutout(requested_bbox, steps, channel_slice)
 
-  def download_point(self, pt, size=256, mip=None):
+  def download_point(self, pt, size=256, mip=None, allow_cropped=False):
     """
     Download to the right of point given in mip 0 coords.
     Useful for quickly visualizing a neuroglancer coordinate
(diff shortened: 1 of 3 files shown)


cloudvolume/cloudvolume.py
```diff
@@ -1032,7 +1033,7 @@ def __getitem__(self, slices):
 
     return self._boss_cutout(requested_bbox, steps, channel_slice)
 
-  def download_point(self, pt, size=256, mip=None):
+  def download_point(self, pt, size=256, mip=None, allow_cropped=False):
```
william-silversmith: CloudVolume already has a mode called "autocrop" that does this. Does it make sense to have a separate parameter for this? Not a rhetorical question.

cloudvolume/cloudvolume.py
```diff
@@ -1055,10 +1056,19 @@ def download_point(self, pt, size=256, mip=None):
...
     saved_mip = self.mip 
-    self.mip = mip 
-    img = self[bbox]
+    self.mip = mip
+    try:
+      img = self[bbox]
+    except ValueError:
+      self.mip = saved_mip
+      print(traceback.format_exc())
+      raise ValueError(
```
william-silversmith: Maybe we should make a new exception like `OutOfBoundsError`. Upside: more clear downside: requires an import from cloudvolume.exceptions to handle.

cloudvolume/cloudvolume.py
```diff
@@ -1055,10 +1056,19 @@ def download_point(self, pt, size=256, mip=None):
...
+      bbox = Bbox.clamp(bbox, self.mip_bounds(mip))
     
     saved_mip = self.mip 
-    self.mip = mip 
-    img = self[bbox]
+    self.mip = mip
+    try:
+      img = self[bbox]
+    except ValueError:
+      self.mip = saved_mip
```
william-silversmith: Very nice. :heart_eyes:

cloudvolume/cloudvolume.py
manuel-castro: I like it

cloudvolume/cloudvolume.py
manuel-castro: Autocrop is cool! Didn't know that was a thing. Don't think we need this flag then. I think I found an autocrop bug though, I'll open an issue

Issue #78, opened 2018-03-16
Storage Dictionary


william-silversmith:
@jabae was thinking of using Storage to upload to his bucket, but the bucket was structured like gs://BUCKET without a dataset or layer. 

We should make this use case more convenient while maintaining neuroglancer related safety.

william-silversmith:
Nacho and Jingpeng mentioned the idea of using a dict interface at various points. Why not make it work like this?

william-silversmith:
The dict idea works well if you only care about binary files. If you want JSON, downloading works well, but uploading is weird (how do you specify JSON?).

nkemnitz:
What exactly was the dictionary idea and where does it come into play with arbitrary storage paths?

william-silversmith:
The dictionary idea was that you could do something like:

```python
stor = Storage("gs://bucket/stairway/to/heaven/")
h5img = stor['temporal.h5']
stor['info'] = x
```
The implementation details aren't related to arbitrary paths, but the ideas are linked in terms of use case as it makes Storage a more convenient way to access any arbitrary cloud files.

william-silversmith:
Resolved in #230

PR #228, opened 2019-06-14
refactor: split up Storage, add components of Green threads

william-silversmith:
Start refactoring the Storage code, introduce GreenStorage without mandating it. Use the job scheduler which is a bit cleaner to read.


william-silversmith:
The dict feature resolves #78

william-silversmith:
A search of the seung-lab organization turned up no uses of the deprecated methods.

PR #231, opened 2019-06-17
fix: microviewer installation, pixel value update, scroll direction

shangmu:
Some microviewer fixes.
The minor fix in the last of the several commits was because the `ext` folder was installing outside the package's own directory due to the `../` path (e.g. into python3.6/site-packages/ext/ ) which is probably unexpected to the user. Feel free to roll back the HEAD if you'd rather keep it as is.


william-silversmith:
I think it is also important to update MANIFEST.in as well. Can you explain why the negative "wheel" value? It might have different effects on different operating systems.

shangmu:
My assumption is scroll up = + while in screen coords downward = + so the direction is wrong. See the `* -0.01` on line 4 here: https://developer.mozilla.org/en-US/docs/Web/API/Element/wheel_event
According to the same doc page it appears the real proper way might be using scroll events instead.

shangmu:
MANIFEST.in is likely not used and broken at the moment anyway: if you do 
`python setup.py sdist` you get 
```warning: LocalManifestMaker: MANIFEST.in, line 1: 'recursive-include' expects <dir> <pattern1> <pattern2> ...```

**Comment by shangmu** [comment]
By the way I forgot to mention, although the doc page was mozilla, the bug was found on chromium

**Comment by william-silversmith** [comment]
I think given the "natural scrolling" on mac touchpads, the negative sign is fairly arbitrary, but I will admit I always get confused for a half-second on my desktop. I need to test the Manifest component but I'm in a weird state at the moment so I'll try this later.

**Comment by shangmu** [comment]
Well since it's the event named "wheel" that we are talking about then either apple or the browser should do whatever necessary to convert touchpad to the wheel convention

**Comment by william-silversmith** [comment]
Looks like now that microviewer is underneath `cloudvolume` the js files are getting included. I moved things around to make more sense.

I tested the scrolling and it does feel better on a desktop. Would you mind checking out my changes? I think yours are good.

**Comment by william-silversmith** [comment]
Actually, I think you might have missed a spot. Check out the `setup_hyper_volume` function on line 157.

**Comment by shangmu** [comment]
Not sure what you mean; they were already included in the distribution.
>  the js files are getting included. 

Fixed hyper volume

**Comment by william-silversmith** [comment]
Previously they were included because of MANIFEST.in, now they are included because they're directly inside the cloudvolume package.

Thanks for this update Shang!

File: cloudvolume/microviewer/__init__.py [hunk 1]
```diff
@@ -12,7 +12,7 @@
 import numpy as np
 from tqdm import tqdm
 
-from .lib import Vec, Bbox, mkdir, save_images, ExtractedPath, yellow
+from cloudvolume.lib import Vec, Bbox, mkdir, save_images, ExtractedPath, yellow
```
shangmu: by the way you could use `..lib` if you prefer

PR #239, opened 2019-07-17
feat: Read Sharded Precomputed Skeletons 

william-silversmith:
This is the easiest one, so it's going first. ^_^


william-silversmith:
Concerned that murmurhash3 has no binaries. Might have to make our own fork...

jbms:
You might find the JavaScript implementation in Neuroglancer informative.

unidesigner:
Just a thought: I could imagine 3 different requests one might want to do:
1. Fetch a single skeleton based on ID
2. Fetch all skeletons given a spatial bounding box
3. Fetch all the skeletons in the entire volume

jbms:
Number 1 and 3 are easy enough with the existing precomputed skeleton
format.  Number 2 would require an additional spatial index.  For the FAFB
volume the total size of the skeletons is not too large so you could just
download them all and then filter them locally.

william-silversmith:
Current test code. This [URL](https://neuroglancer-demo.appspot.com/fafb.html#!%7B%22layers%22:%5B%7B%22source%22:%22precomputed://gs://neuroglancer-fafb-data/fafb_v14/fafb_v14_orig_sharded%22%2C%22type%22:%22image%22%2C%22name%22:%22fafb_v14%22%2C%22visible%22:false%7D%2C%7B%22source%22:%22precomputed://gs://neuroglancer-fafb-data/fafb_v14/fafb_v14_clahe_sharded%22%2C%22type%22:%22image%22%2C%22name%22:%22fafb_v14_clahe%22%7D%2C%7B%22source%22:%22precomputed://gs://neuroglancer-fafb-data/elmr-data/FAFBNP.surf%22%2C%22type%22:%22segmentation%22%2C%22segments%22:%5B%221%22%2C%2210%22%2C%2211%22%2C%2212%22%2C%2213%22%2C%2214%22%2C%2215%22%2C%2216%22%2C%2217%22%2C%2218%22%2C%2219%22%2C%222%22%2C%2220%22%2C%2221%22%2C%2222%22%2C%2223%22%2C%2224%22%2C%2225%22%2C%2226%22%2C%2227%22%2C%2228%22%2C%2229%22%2C%223%22%2C%2230%22%2C%2231%22%2C%2232%22%2C%2233%22%2C%2234%22%2C%2235%22%2C%2236%22%2C%2237%22%2C%2238%22%2C%2239%22%2C%224%22%2C%2240%22%2C%2241%22%2C%2242%22%2C%2243%22%2C%2244%22%2C%2245%22%2C%2246%22%2C%2247%22%2C%2248%22%2C%2249%22%2C%225%22%2C%2250%22%2C%2251%22%2C%2252%22%2C%2253%22%2C%2254%22%2C%2255%22%2C%2256%22%2C%2257%22%2C%2258%22%2C%2259%22%2C%226%22%2C%2260%22%2C%2261%22%2C%2262%22%2C%2263%22%2C%2264%22%2C%2265%22%2C%2266%22%2C%2267%22%2C%2268%22%2C%2269%22%2C%227%22%2C%2270%22%2C%2271%22%2C%2272%22%2C%2273%22%2C%2274%22%2C%2275%22%2C%228%22%2C%229%22%5D%2C%22skeletonRendering%22:%7B%22mode2d%22:%22lines_and_points%22%2C%22mode3d%22:%22lines%22%7D%2C%22name%22:%22neuropil-regions-surface%22%2C%22visible%22:false%7D%2C%7B%22source%22:%22precomputed://gs://fafb-ffn1-20190521/segmentation%22%2C%22type%22:%22segmentation%22%2C%22segments%22:%5B%221822975381%22%2C%222016390195%22%5D%2C%22skeletonRendering%22:%7B%22mode2d%22:%22lines_and_points%22%2C%22mode3d%22:%22lines%22%7D%2C%22name%22:%22fafb-ffn1-20190521%22%2C%22visible%22:false%7D%2C%7B%22source%22:%22precomputed://gs://fafb-ffn1-20190521/segmentation%22%2C%22type%22:%22segmentation%22%2C%22skeletons%22:%22precomputed://gs://fafb-ffn1-20190521/segmentation/skeletons_32nm%22%2C%22segments%22:%5B%221822975381%22%2C%222016390195%22%5D%2C%22skeletonRendering%22:%7B%22mode2d%22:%22lines_and_points%22%2C%22mode3d%22:%22lines%22%7D%2C%22name%22:%22fafb-ffn1-20190521-skeletons%22%7D%5D%2C%22navigation%22:%7B%22pose%22:%7B%22position%22:%7B%22voxelSize%22:%5B4%2C4%2C40%5D%2C%22voxelCoordinates%22:%5B142190.515625%2C64282.38671875%2C3130.26416015625%5D%7D%7D%2C%22zoomFactor%22:10.631667159656855%7D%2C%22perspectiveOrientation%22:%5B0.2667730450630188%2C-0.6194830536842346%2C-0.6498712301254272%2C0.35034310817718506%5D%2C%22perspectiveZoom%22:105.04864808962009%2C%22showSlices%22:false%2C%22selectedLayer%22:%7B%22layer%22:%22fafb-ffn1-20190521%22%7D%2C%22layout%22:%223d%22%7D) is sneaky (skeletons isn't specified in the info file).
```python
from cloudvolume import CloudVolume
import numpy as np

from cloudvolume.datasource.precomputed.skeleton.sharded import ShardedPrecomputedSkeletonSource

cv = CloudVolume('precomputed://gs://fafb-ffn1-20190521/segmentation', use_https=True)
cv.meta.info['skeletons'] = 'skeletons_32nm'
cv.skeleton.meta.refresh_info()
cv.skeleton = ShardedPrecomputedSkeletonSource(cv.skeleton.meta, cv.cache, cv.config)
s1 = cv.skeleton.get(1822975381)
```

william-silversmith:
Just saw your comments @jbms. I'll address them tomorrow.

william-silversmith:
I was testing against the URL above and found that while the spec calls for "murmurhash3_x86_128" (all lowercase) the actual value in the wild was "murmurhash3_X86_128" (uppercase X).

jbms:
I corrected those info files, and will make sure to generate using "murmurhash3_x86_128" in the future.  Neuroglancer uses case-insensitive comparison so it wasn't noticed before.

william-silversmith:
It looks like https://storage.googleapis.com/fafb-ffn1-20190521/segmentation/info still doesn't list the skeletons directory. Is that something you can fix? It might be worth accommodating this use case in my code regardless.

jbms:
The new mesh/skeleton formats have their own info file, with the intention that you could just use a URL like https://storage.googleapis.com/fafb-ffn1-20190521/segmentation/skeletons_32nm to refer to the skeletons.

william-silversmith:
@jbms I seem to be having trouble getting my hash function to output the same thing as neuroglancer. Am I doing this right?

jbms:
I think when you were using the mmh3 library you were first converting the key to a base 10 string representation and then hashing that.  Instead, you should convert the key to its 8-byte little endian representation and hash that.

william-silversmith:
I gave it a try using mmh3.hash64(np.uint64(x)) as well and failed to get the right answer using either half of the result. It's a little frustrating because I'm trying to get the code to output "42" as the shard number from 1822975381, which is ironically funny.

william-silversmith:
code: chunkid, shard_number

`uint64(mmh3.hash64(uint64(x))[0])`: 8608683511501517555 6e
`uint64(mmh3.hash64(uint64(x))[1])`: 18148385391598999711 33
`uint64(mmh3.hash128(uint64(x)) & uint64(0x00000000ffffffff))`: 4275008243 6e
`uint64(mmh3.hash128(uint64(x)) & uint64(0xffffffff00000000))`: 8608683507226509312 00
`uint64(uint64(mmh3.hash128(uint64(x)) & uint64(0xffffffff00000000)) >> uint64(32))`: 2004365322 06
`murmurHash3_x86_128Hash64Bits`: 9330517092871124057 66

jbms:
It looks like you need to pass x64arch=False to get the x86 variant.  There are two variants of MurmurHash3 128 and they don't produce the same output.

william-silversmith:
Thank the maker! That did it. ^_^ Thank you Jeremey!

william-silversmith:
While this PR needs cleanup, this is now able to read skeletons. (!!!)

william-silversmith:
Remaining tasks that come to mind. Any others?

- [ ] add tests
- [ ] skeletons use info file attributes and transforms
- [ ] fix PrecomputedSkeletonSource.from_cloudpath

Right now I'm only caching index files as they are the only entire file I download. I could cache individual skeletons downloaded or download entire `.data` files. It might be worth a configuration option or a separate call.  

In the future, we'll want to find ways to condense and pipeline requests for higher efficiency.

jbms:
Neuroglancer caches individual minishard indexes, but not entire shard index files.  If there is locality in the ids accessed (particularly for the sharded volume format), caching the minishard indexes is particularly useful.  Caching the entire shard index may be less useful since even if you get a cache hit you still have to fetch the minishard index.

william-silversmith:
Four different ways of caching:

1) Save only the minishard index: Speeds up request 1 of 3
2) Save downloaded skeletons: Convert 3 requests into 1, but recreates the original fragmentation problem locally (fine for sparse usage)
3) Save everything: Good for batch processing large amounts of the data repeatedly.
4) mmap `.data` files on disk and only write the sections we actually download. Requires a sentinel value (all zeros for skeletons?) or additional dirty bit to know when to request from remote.

jbms:
For caching I was thinking in the context of in-memory caching, not disk caching.  I see that for disk caching there are some different considerations.

To be clear, there is the "shard index", which is in the .index file, and specifies the location within the .data file of the "minishard index" for each minishard, and then the "minishard index", which specifies the location of each entry (i.e. skeleton).  If you cache the minishard index (and cache it by (shard,minishard), not by location within the .data file), then a subsequent request for another skeleton that happens to be in the same minishard only requires 1 additional request (to download the actual skeleton data), since you already know its location from the cached minishard index.

If you are worried about storing too many small files on the local filesystem, you could use an embedded database like LMDB or SQLite.  LMDB uses mmap internally and is quite fast.

william-silversmith:
I think for now I'm not going to worry too much about doing anything more complex for now. I'm going to cache skeletons sparsely as they come in, which is very performant and probably conform's to most people's naive expectation of what caching would do. I'll also cache the index files. We can work on more complex modes as use cases appear.

william-silversmith:
Debating... should skeletons be returned in raw coordinates or in transformed coordinates? What will people expect? Either way, there should be a method to call to switch between them.

unidesigner:
I'd expect to get the skeleton coordinates transformed, i.e. in nm units - and with a switch to get them in voxel units. This is also how it is done in CATMAID. It generally simplifies all downstream skeleton-based morphology analysis code.

william-silversmith:
Awesome. I'll do it that way then. Thanks Stephan!

william-silversmith:
It's the segment id (or numerical label) that corresponds to a Neuroglancer segmentation.

unidesigner:
> It's the segment id (or numerical label) that corresponds to a Neuroglancer segmentation.

Got it. Is there any reason this must be a single Int, and not a list of Ints? (e.g. for skeletons derived from multiple joined segments)

william-silversmith:
You can use it that way but some functions (such as upload) will break.
It's not a meaningful identifier in neuroglancer.

william-silversmith:
Okay, I think this is almost ready for release. We just need a few more tests.

william-silversmith:
I spoke too soon.

william-silversmith:
:tada:

Issue #247, opened 2019-07-25
use_https not working


fcollman:
0.56.1 appears to have a problem with properly passing the use_https flag on to all the subsequent cloudvolume plugins for graphene, making it impossible to access public cloud buckets that are used by graphene without google credentials.

```---------------------------------------------------------------------------
Empty                                     Traceback (most recent call last)
/pipeline/anaconda/envs/celltypes/lib/python3.7/site-packages/cloudvolume/connectionpools.py in get_connection(self)
     51       try:
---> 52         conn = self.pool.get(block=False)
     53         self.pool.task_done()

/pipeline/anaconda/envs/celltypes/lib/python3.7/queue.py in get(self, block, timeout)
    166                 if not self._qsize():
--> 167                     raise Empty
    168             elif timeout is None:

Empty: 

During handling of the above exception, another exception occurred:

DefaultCredentialsError                   Traceback (most recent call last)
<ipython-input-56-647de00127a1> in <module>
      4     print(k)
      5     seg_id = int(row.pt_root_id)
----> 6     mesh = mm.mesh(seg_id=seg_id)
      7     sk_file = os.path.join(skel_dir, f'{seg_id}.h5')
      8     sk = skeleton_io.read_skeleton_h5(sk_file)

/pipeline/anaconda/envs/celltypes/lib/python3.7/site-packages/meshparty/trimesh_io.py in mesh(self, filename, seg_id, cache_mesh, merge_large_components, stitch_mesh_chunks, overwrite_merge_large_components, force_download)
    254         """
    255         assert filename is not None or \
--> 256             (seg_id is not None and self.cv is not None)
    257 
    258         if filename is not None:

/pipeline/anaconda/envs/celltypes/lib/python3.7/site-packages/meshparty/trimesh_io.py in cv(self)
    221         if self._cv is None and self.cv_path is not None:
    222             self._cv = cloudvolume.CloudVolume(self.cv_path, parallel=10,
--> 223                                                use_https=self._map_gs_to_https)
    224 
    225         return self._cv

/pipeline/anaconda/envs/celltypes/lib/python3.7/site-packages/cloudvolume/cloudvolume.py in __new__(cls, cloudpath, mip, bounded, autocrop, fill_missing, cache, compress_cache, cdn_cache, progress, info, provenance, compress, non_aligned_writes, parallel, delete_black_uploads, green_threads, use_https)
    185     path = strict_extract(cloudpath)
[...]

PR #248, opened 2019-07-25
fix: added use_https to graphene metadata service

fcollman:
I think i found the bug related to #247 , the GrapheneMetaDataService needed to allow remapping of https on the data_dir.  Please let me know if this is the correct way to fix this within the plugin architecture.

### DIFF
--- a/cloudvolume/cloudvolume.py
+++ b/cloudvolume/cloudvolume.py
@@ -180,7 +180,6 @@ class CloudVolume(object):
 
     kwargs = locals()
     del kwargs['cls']
-    del kwargs['use_https']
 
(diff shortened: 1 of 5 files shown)


william-silversmith:
This is good. The only comment I'd make is that it would be been better to abstract the functionality. In the future, we'll need to remember to update two places in the code.

william-silversmith:
Looks like this change also conflicts with Precomputed. I added a really specific fix for it.

william-silversmith:
This might actually be the right answer... make all subclasses accept use_https since we want to enable support for all subclasses anyway.

Issue #251, opened 2019-07-30
graphene cutout service bug


fcollman:
```Python
File "/Users/forrestc/ConnectomeStack/EM_mesh_analysis/em_mesh_analysis/Fix_mesh.py", line 133, in get_fixed_seg_mask
    id_list]
  File "/Users/forrestc/anaconda3/envs/celltypes/lib/python3.7/site-packages/cloud_volume-0.56.1-py3.7.egg/cloudvolume/frontends/graphene.py", line 96, in __getitem__
    root_ids = list(slices.pop())
AttributeError: 'tuple' object has no attribute 'pop'
```

Not sure if i'm just using it wrong, but putting the 4th dimension as a set of IDs to lookup appears to break the cloud-volume graphene cutout front end. 

Here is the code that was calling it 

```Python
        id_list = [nuc_seg_id]
        if cell_seg_id is not None:
            id_list.append(cell_seg_id)
        seg_cutout = seg_vol[mins[0]:maxs[0],
                             mins[1]:maxs[1],
                             mins[2]:maxs[2],
                             id_list]
```

fcollman:
I think there is something further wrong with this code... on the graphene_test branch #254 I discovered this odd difference

```Python
In [19]: cv.download(bbox, mask_base=True, root_ids=[648518346349523869])                                                                  
Downloading:   0%|                                                                                                   | 0/1 [00:00<?, ?it/s]
Out[19]: 
array([[[[648518346349523869]],

        [[648518346349523869]],

        [[648518346349523869]]],


       [[[648518346349523869]],

        [[648518346349523869]],

        [[648518346349523869]]],


       [[[648518346349523869]],

        [[648518346349523869]],

        [[648518346349523869]]],


       [[[648518346349523869]],

        [[648518346349523869]],

        [[648518346349523869]]]], dtype=uint64)

In [20]: bbox                                                                                                                              
Out[20]: Bbox([54314, 22606, 1136],[54318, 22609, 1137], dtype=int32)

In [21]: cv[54314, 22606, 1136,[648518346349523869]]                                                                                       
Downloading:   0%|                                                                                                   | 0/1 [00:00<?, ?it/s]
Out[21]: array([[[0]]], dtype=uint64)
```

fcollman:
removing some try/excepts i found this 

```Python 
Traceback (most recent call last):
  File "test_slices.py", line 18, in <module>
    print(cv[54314, 22606, 1136,[648518346349523869]])
[...]

PR #254, opened 2019-07-30
graphene tests and cutout bug fix

fcollman:
This is a draft aimed to address #251 but also start to have some test infrastructure surrounding graphene with a properly mocked out PCG.  Submitting a PR as a draft to stimulate discussion.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,8 @@ install:
 - if [[ $PYTHON_MAJOR_VERSION == 3 ]]; then pip3 install numpy; fi
 - if [[ $PYTHON_MAJOR_VERSION == 2 ]]; then pip install fpzip; fi
 - if [[ $PYTHON_MAJOR_VERSION == 3 ]]; then pip3 install fpzip; fi
-- if [[ $PYTHON_MAJOR_VERSION == 2 ]]; then pip install -e .; fi
-- if [[ $PYTHON_MAJOR_VERSION == 3 ]]; then pip3 install -e .; fi
+- if [[ $PYTHON_MAJOR_VERSION == 2 ]]; then pip install -e .[test]; fi
+- if [[ $PYTHON_MAJOR_VERSION == 3 ]]; then pip3 install -e .[test]; fi
 - if [[ $PYTHON_MAJOR_VERSION == 2 ]]; then python setup.py sdist; fi
 - if [[ $PYTHON_MAJOR_VERSION == 3 ]]; then python3 setup.py sdist bdist_wheel --universal; fi
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,8 @@ install:
 - if [[ $PYTHON_MAJOR_VERSION == 3 ]]; then pip3 install numpy; fi
 - if [[ $PYTHON_MAJOR_VERSION == 2 ]]; then pip install fpzip; fi
 - if [[ $PYTHON_MAJOR_VERSION == 3 ]]; then pip3 install fpzip; fi
-- if [[ $PYTHON_MAJOR_VERSION == 2 ]]; then pip install -e .[test,boss]; fi
-- if [[ $PYTHON_MAJOR_VERSION == 3 ]]; then pip3 install -e .[test,boss]; fi
+- if [[ $PYTHON_MAJOR_VERSION == 2 ]]; then pip install -e .[boss,test]; fi
+- if [[ $PYTHON_MAJOR_VERSION == 3 ]]; then pip3 install -e .[boss,test]; fi
 - if [[ $PYTHON_MAJOR_VERSION == 2 ]]; then python setup.py sdist; fi
 - if [[ $PYTHON_MAJOR_VERSION == 3 ]]; then python3 setup.py sdist bdist_wheel --universal; fi
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,8 @@ install:
 - if [[ $PYTHON_MAJOR_VERSION == 3 ]]; then pip3 install numpy; fi
 - if [[ $PYTHON_MAJOR_VERSION == 2 ]]; then pip install fpzip; fi
 - if [[ $PYTHON_MAJOR_VERSION == 3 ]]; then pip3 install fpzip; fi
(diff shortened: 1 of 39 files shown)


fcollman:
I found a number of other bugs in the graphene cutout service that I think are now fixed.. the most important being that bbox's to get_leaves should be passed as mip=0.  I personally think we should change to allowing the user to specify cutouts in the mip they request, and then translate that bbox to the mip 0 level and pass to get_leaves.  That will make it unified with the precomputed interface.

cloudvolume/frontends/graphene.py
```diff
@@ -92,17 +91,19 @@ def __getitem__(self, slices):
...
+      slices = list(slices)
+      root_ids = list(toiter(slices.pop()))
       return self.download(
         slices, mip=self.mip, 
         parallel=self.parallel, root_ids=root_ids
       )
     except TypeError: 
       # The end of the array was not iterable, 
       # and thus not the root ids.
+      print('warning no root ids selected, downloading supervoxels')
```
william-silversmith: Not sure this except clause serves any purpose. It was guarding the access of [-1] previously, but that is no longer true. We are now allowing the last item to be an int instead of a list as well.

.travis.yml
```diff
@@ -19,8 +19,8 @@ install:
 - if [[ $PYTHON_MAJOR_VERSION == 3 ]]; then pip3 install numpy; fi
 - if [[ $PYTHON_MAJOR_VERSION == 2 ]]; then pip install fpzip; fi
 - if [[ $PYTHON_MAJOR_VERSION == 3 ]]; then pip3 install fpzip; fi
-- if [[ $PYTHON_MAJOR_VERSION == 2 ]]; then pip install -e .; fi
-- if [[ $PYTHON_MAJOR_VERSION == 3 ]]; then pip3 install -e .; fi
+- if [[ $PYTHON_MAJOR_VERSION == 2 ]]; then pip install -e .[test]; fi
```
william-silversmith: Should be .[test,boss]

cloudvolume/frontends/graphene.py
fcollman: What happens if someone only passes 3 slices?

.travis.yml
fcollman: fixed

cloudvolume/frontends/graphene.py
fcollman: i tried an alternative approach.. so now it acts like precomputed giving  you straight supervoxels if you slice it normally.. and you pass rootIDs to get remapping behavior.

william-silversmith: If someone passes fewer than four slices, the rest are filled with implied `:`. The behavior you described seems sensible, though if someone passes in an programatically generated empty list they will be surprised by the result.

fcollman: what do you think of the new option?

fcollman: which tests for len(slices)==4 and does different things depending

william-silversmith: Hmm, I feel like that makes the interface a little inconvenient. 

I think you'll probably want the following interface?

```python
ws_img = cv[:]    # watershed
ws_img = cv[:,:]  # watershed
ws_img = cv[:,:,:] # still watershed

mapped = cv[:, []] # maps to black
mapped = cv[:, 0] # maps to black
mapped = cv[:,:, [ 5 ]] # agglomerates 5, else black
```

fcollman: I'm not sure that follows normal numpy conventions.    I think if you want to index the 4th axis, you always need to put it in the 4th spot.  Specifying fewer slices than axis is fine, and you default to 'all' of them, but cv[:, []] would be an empty array, cv[:, 0]  would be the all rows, first column, all z, supervoxels. cv[:,:, [ 5 ]]  is z=5. as would mapped = cv[:,:, 5].

william-silversmith: I guess I wasn't seeing the root ids as a distinct axis but as something else tacked on the end. It does make sense to see it as a kind of axis. It does mean you lose some of these implicit niceties. I'm okay with it if you are.

fcollman: i do think a reasonable option would be to say.. no the __getitem__ interface only gives you slicing into the precomputed format and the download method has the root_ids option.  It's true that there won't be code that easily translates between graphene and precomputed anyway, and you could give precomputed a download method too that makes it behave the same way when you pass rootIds and mask_base=True.  What do you think?

william-silversmith: I actually quite like this idea to make `__getitem__` behave in a more expected way. The addition of non-slices to it is a kludge that we came up with before the download method existed. I'm not sure it makes sense to backport it to Precomputed. The masking operation is essential to usage of Graphene, but not so much for Precomputed (also the terminology of root_ids doesn't make sense there). It's easy enough to use the `fastremap` module to perform these operations in an explicit way.

If there's a clear use case for backporting I could be convinced otherwise though. I want to keep the interfaces simple, but not so simple they impede people.

fcollman: then i think that's also an argument for making the default behavior of graphene to leave the supervoxel IDs in place, because you'd like downstream analysis code to work reasonably similarly on both sources.  Having black masks all around is very different than be surrounding by other IDs.  Especially considering that id=0 is a special case that means . something different.. i.e. no segmentation.

cloudvolume/datasource/graphene/__init__.py
```diff
@@ -49,7 +50,10 @@ def create_graphene(
     )
 
     mesh = GrapheneMeshSource(meta, cache, config)
-    # skeleton = PrecomputedSkeletonSource(meta, cache, config)
+    try:
```
william-silversmith: This a little over-defensive imo. If Skeletons are crashing because of the dang info file, 1) I want to know about it 2) You're probably going to see issues elsewhere 3) This only covers 1 of 5 potential protocols.

william-silversmith: I take this back slightly, I still feel this is overly defensive because errors should be handled by the skeleton source. However, graphene really only uses the https protocol.

william-silversmith:
@fcollman This should do it! Give it a try.

PR #261, opened 2019-08-19
feat: support for chunked graph server authentication

akhileshh:
* store credentials in file named `chunkedgraph-secret.json`, eg: `{"token":"<auth_token>"}`
* token will be in `auth_header` dictionary in meta class

### DIFF
--- a/cloudvolume/datasource/graphene/metadata.py
+++ b/cloudvolume/datasource/graphene/metadata.py
@@ -11,14 +11,19 @@ from ...secrets import chunkedgraph_credentials
 from ..precomputed import PrecomputedMetadata
 
 class GrapheneMetadata(PrecomputedMetadata):
-  def __init__(self, cloudpath, use_https=False, *args, **kwargs):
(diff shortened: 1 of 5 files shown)


cloudvolume/datasource/graphene/metadata.py
```diff
@@ -7,20 +7,26 @@
 
 from ... import exceptions
 from ... import paths
+from ...secrets import chunkedgraph_credentials
 from ..precomputed import PrecomputedMetadata
 
 class GrapheneMetadata(PrecomputedMetadata):
   def __init__(self, cloudpath, use_https=False, *args, **kwargs):
```
william-silversmith: The GrapheneMetadata class should also support providing the token directly (including blanking it). It should by default use the saved value. This will make it easier to support remixing components.

akhileshh: done

william-silversmith:
Thank you @akhileshh!

PR #270, opened 2019-09-27
feat: add redirection feature to info files

william-silversmith:
We were discussing a problem we are having where some of our collaborators have spreadsheets of neuroglancer links, but we are moving the layer they point to into cheaper storage. We are thinking of instrumenting Neuroglancer and CloudVolume with a redirect capability. 

I looked around for HTTP 301 support on GCS, but it didn't seem to exist, so we are proposing to handle redirects like so:

info:
```json
{
    ... may contain additional keys ... 
    "redirect": "precomputed://gs://new/location/"
}
```
The redirect process would abort with an error once a cycle is detected. 

This adds the `max_redirects` flag to CloudVolume so that you can specify how many redirects are allowed. 0 for direct access, 10 by default.

If a redirect occurs, a readonly flag is set on image upload automatically, but can be disabled.

```python
# fixing a broken redirect
vol = CloudVolume(..., max_redirects=0)
del vol.info['redirect']
vol.commit_info()

# setting images to writable 
vol = CloudVolume(...) # redirected 1 or more times
vol.image.readonly = False
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -338,7 +338,8 @@ CloudVolume(cloudpath,
      cache=False, compress_cache=None, cdn_cache=False, progress=INTERACTIVE, info=None, provenance=None, 
      compress=None, non_aligned_writes=False, parallel=1,
      delete_black_uploads=False, background_color=0, 
-     green_threads=False)
+     green_threads=False, use_https=False,
+     max_redirects=10)
 ```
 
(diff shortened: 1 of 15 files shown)


william-silversmith:
Should there be a warning printed on redirection?

PR #271, opened 2019-10-04
sets ACL by default for s3 interface

falkben:
On S3 when a user from a separate account puts objects into a S3 bucket that they do not own, the default behavior is that the actual owner of that bucket has no privileges to those objects.  This is in my experience rarely desired behavior and can be quite tedious to fix after the fact.

This PR attempts to set a default ACL for objects created using cloud volume.

This is not generalized and might be necessary for other clouds as well.

Also, would need to be propagated through the code in order for a user to be able to supply a custom ACL, in the case that they actually want ACL to limit owner control.

Happy to to more work on this as needed but wanted to check in first to make sure it was something wanted or if you had some other idea how to tackle this.

### DIFF
--- a/cloudvolume/storage/storage_interfaces.py
+++ b/cloudvolume/storage/storage_interfaces.py
@@ -338,7 +338,7 @@ class S3Interface(StorageInterface):
     return posixpath.join(self._path.no_bucket_basepath, self._path.layer, file_path)
 
   @retry
-  def put_file(self, file_path, content, content_type, compress, cache_control=None):
+  def put_file(self, file_path, content, content_type, compress, cache_control=None, ACL="bucket-owner-full-control"):
     key = self.get_path_to_file(file_path)
 
     attrs = {
@@ -346,6 +346,7 @@ class S3Interface(StorageInterface):
       'Body': content,
       'Key': key,
       'ContentType': (content_type or 'application/octet-stream'),
+      'ACL': ACL,
     }
 
     if compress:


william-silversmith:
This is cool. It makes sense to be able to set this per a file. I would like to make setting permissions uniform across all storage objects, but I'm not sure how to do that yet. Would using `permissions` instead of `ACL` be okay? That way we could add it to multiple storage types.

falkben:
By `permissions` do you mean the secrets file?  If so, then I think that could work (or environment variable)

For usability, it might be nice to have a kwargs input so that a user could specify whatever additional parameters they like in a dictionary and have them applied at the upload step.

In my case, I actually switched the ACL from what is in this PR ("bucket-owner-full-control") for default to "public-read" (which also sets the owner to the bucket owner) but makes the objects public. Not implying that public-read should be the default, just that it is nice to be able to easily change that option. 

Using cloud-volume across AWS accounts is a bit tricky right now, but would be very helpful in the case of an [AWS Public Data](https://aws.amazon.com/opendata/) bucket (they provide storage but no compute)

william-silversmith:
I meant only that you change the keyword ACL to permissions since ACL only
applies to S3. The function interface should be uniform so that we can pass
in keywords from the top level later on.

CloudVolume can be used across AWS accounts can be made pretty easy by
either setting per-bucket secrets files (for writing) or using the https
interface for reading (use_https=True) if the bucket is set to public read.

falkben:
Oh, yes, the param name could be whatever you like it to be.  Google cloud also uses [ACL](https://cloud.google.com/storage/docs/access-control/lists) FYI

I have used CloudVolume across accounts (using environment variables to load in secrets) for writing.  DIdn't mean to imply it wasn't possible.  However, I did have to account for ACL so the owner doesn't get locked out of being able to modify the files on their own bucket. :roll_eyes:

william-silversmith:
I'll merge this for now, but the keyword might change in the future.

PR #278, opened 2019-10-21
feat: add brotli compression option

falkben:
This adds brotli compression with a default compression level of 6.

My [testing](https://nbviewer.jupyter.org/github/neurodata/boss-export/blob/lambda/scripts/bench_compress.ipynb) has shown that there's a nice 10% compression rate advantage, with a small cost in compression/decompression speed.

I added tests for brotli with localfile volumes, but wasn't able to run the s3/gcs specific tests.

I did test "pulling" one of our brotli compressed datasets from a public precomputed test volume we have which you should be able to test as well.  See below:

```python3
import os

import boto3
import numpy as np
from PIL import Image

from cloudvolume import CloudVolume

path = "s3://open-neurodata-test/ZBrain/ZBrain/ZBB_y385-Cre"
vol = CloudVolume(path)

image = vol[0:512, 0:512, 0:16]

img = Image.fromarray(image[:,:,0,0])
img.save("/tmp/removeme/img.tif")
```

### DIFF
--- a/cloudvolume/storage/storage_interfaces.py
+++ b/cloudvolume/storage/storage_interfaces.py
@@ -190,10 +190,8 @@ class GoogleCloudStorageInterface(StorageInterface):
     key = self.get_path_to_file(file_path)
     blob = self._bucket.blob( key )
 
-    # keep default gzip
-    if compress == "br":
-      blob.content_encoding = "br"
-    elif compress:
+    # gcloud disable brotli until content-encoding works
+    if compress:
       blob.content_encoding = "gzip"
     if cache_control:
       blob.cache_control = cache_control
--- a/cloudvolume/storage/storage_interfaces.py
+++ b/cloudvolume/storage/storage_interfaces.py
@@ -15,6 +15,7 @@ import tenacity
 
 from cloudvolume.connectionpools import S3ConnectionPool, GCloudBucketPool
 from cloudvolume.lib import mkdir
+from cloudvolume.exceptions import UnsupportedCompressionType
 
 # This is just to support pooling by bucket
 class keydefaultdict(defaultdict):
@@ -115,7 +116,7 @@ class FileInterface(StorageInterface):
 
   def exists(self, file_path):
     path = self.get_path_to_file(file_path)
-    return os.path.exists(path) or os.path.exists(path + '.gz')
+    return os.path.exists(path) or os.path.exists(path + '.gz') or os.path.exists(path + ".br")
 
   def files_exist(self, file_paths):
     return { path: self.exists(path) for path in file_paths }
@@ -126,6 +127,8 @@ class FileInterface(StorageInterface):
       os.remove(path)
     elif os.path.exists(path + '.gz'):
       os.remove(path + '.gz')
+    elif os.path.exists(path + ".br"):
+      os.remove(path + ".br")
 
   def delete_files(self, file_paths):
     for path in file_paths:
@@ -165,14 +168,14 @@ class FileInterface(StorageInterface):
         for filename in files:
           filenames.append(filename)
     
-    def stripgz(fname):
+    def stripext(fname):
       (base, ext) = os.path.splitext(fname)
-      if ext == '.gz':
+      if ext == '.gz' or ext == ".br":
         return base
       else:
         return fname
 
-    filenames = list(map(stripgz, filenames))
+    filenames = list(map(stripext, filenames))
     return _radix_sort(filenames).__iter__()
 
 class GoogleCloudStorageInterface(StorageInterface):
@@ -192,7 +195,7 @@ class GoogleCloudStorageInterface(StorageInterface):
 
     # gcloud disable brotli until content-encoding works
     if compress == "br":
-      raise TypeError("Brotli unsupported on google cloud storage")
+      raise UnsupportedCompressionType("Brotli unsupported on google cloud storage")
     elif compress:
       blob.content_encoding = "gzip"
     if cache_control:
--- a/cloudvolume/storage/storage_interfaces.py
+++ b/cloudvolume/storage/storage_interfaces.py
@@ -191,7 +191,9 @@ class GoogleCloudStorageInterface(StorageInterface):
     blob = self._bucket.blob( key )
 
     # gcloud disable brotli until content-encoding works
-    if compress:
+    if compress == "br":
+      raise TypeError("Brotli unsupported on google cloud storage")
+    elif compress:
       blob.content_encoding = "gzip"
     if cache_control:
       blob.cache_control = cache_control
--- a/cloudvolume/storage/storage_interfaces.py
+++ b/cloudvolume/storage/storage_interfaces.py
@@ -319,6 +319,8 @@ class HttpInterface(StorageInterface):
     else:
       resp = requests.get(key)
       if resp.headers.get('Content-Encoding') == 'br':
+        # needed until requests natively supports brotli
+        # https://github.com/psf/requests/issues/4525
         resp._content = brotli.decompress(resp.content)
     if resp.status_code in (404, 403):
       return None, None
--- a/cloudvolume/storage/storage_interfaces.py
+++ b/cloudvolume/storage/storage_interfaces.py
@@ -67,7 +67,10 @@ class FileInterface(StorageInterface):
     path = self.get_path_to_file(file_path)
     mkdir(os.path.dirname(path))
 
-    if compress:
+    # keep default as gzip
+    if compress == "br":
+      path += ".br"
+    elif compress:
       path += '.gz'
 
     if content \
@@ -87,13 +90,15 @@ class FileInterface(StorageInterface):
   def get_file(self, file_path, start=None, end=None):
     path = self.get_path_to_file(file_path)
 
-    compressed = os.path.exists(path + '.gz')
-      
-    if compressed:
+    if os.path.exists(path + '.gz'):
+      encoding = "gzip"
       path += '.gz'
+    elif os.path.exists(path + '.br'):
+      encoding = "br"
+      path += ".br"
+    else:
+      encoding = None
 
-    encoding = 'gzip' if compressed else None
-    
     try:
       with open(path, 'rb') as f:
         if start is not None:
@@ -184,7 +189,11 @@ class GoogleCloudStorageInterface(StorageInterface):
   def put_file(self, file_path, content, content_type, compress, cache_control=None):
     key = self.get_path_to_file(file_path)
     blob = self._bucket.blob( key )
-    if compress:
+
+    # keep default gzip
+    if compress == "br":
+      blob.content_encoding = "br"
+    elif compress:
       blob.content_encoding = "gzip"
     if cache_control:
       blob.cache_control = cache_control
@@ -349,7 +358,10 @@ class S3Interface(StorageInterface):
       'ACL': ACL,
     }
 
-    if compress:
+    # keep gzip as default
+    if compress == "br":
+      attrs['ContentEncoding'] = 'br'
+    elif compress:
       attrs['ContentEncoding'] = 'gzip'
     if cache_control:
       attrs['CacheControl'] = cache_control
--- a/cloudvolume/storage/storage_interfaces.py
+++ b/cloudvolume/storage/storage_interfaces.py
@@ -318,6 +318,8 @@ class HttpInterface(StorageInterface):
       resp = requests.get(key, headers=headers)
     else:
       resp = requests.get(key)
+      if resp.headers.get('Content-Encoding') == 'br':
+        resp._content = brotli.decompress(resp.content)
     if resp.status_code in (404, 403):
       return None, None
     resp.raise_for_status()
--- a/README.md
+++ b/README.md
@@ -352,7 +352,7 @@ CloudVolume(cloudpath,
 * progress - Show progress bars. Defaults to True if in python interactive mode else default False.
 * info - Use this info object rather than pulling from the cloud (useful for creating new layers).
 * provenance - Use this object as the provenance file.
-* compress - None or 'gzip', force this compression algorithm to be used for upload
+* compress - None, 'gzip' or 'br', force this compression algorithm to be used for upload
 * non_aligned_writes - True/False. If False, non-chunk-aligned writes will trigger an error with a helpful message. If True,
     Non-aligned writes will proceed. Be careful, non-aligned writes are wasteful in memory and bandwidth, and in a mulitprocessing environment, are subject to an ugly race condition. (c.f. https://github.com/seung-lab/cloud-volume/wiki/Advanced-Topic:-Non-Aligned-Writes)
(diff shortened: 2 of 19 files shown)


william-silversmith:
For other readers: Tested Ben's link. This is neuroglancer compatible b/c brotli is DEFLATE compatible.

cloudvolume/storage/storage_interfaces.py
```diff
@@ -87,13 +90,15 @@ def put_file(
   def get_file(self, file_path, start=None, end=None):
     path = self.get_path_to_file(file_path)
 
-    compressed = os.path.exists(path + '.gz')
-      
-    if compressed:
+    if os.path.exists(path + '.gz'):
+      encoding = "gzip"
       path += '.gz'
+    elif os.path.exists(path + '.br'):
```
william-silversmith: Is there an advantage to using the brotli decompressor vs the gzip decompressor? My understanding is that cross compatibility was supposed to be brotli's big contribution. If the difference is slight, it might make sense to just write all these files as .gz to avoid adding an additional syscall + FS access here.

william-silversmith:
This didn't come through in my other comments, but this is a really great contribution! Thank you!

cloudvolume/storage/storage_interfaces.py
william-silversmith: Since it is very uncommon for datasets to have mixed compression types, maybe it would be a good idea to cache the result of the first successful operation?

falkben: They will need to have a different extension from gzip, as they need a different decompressor, and `.br` seems to be the standard.

Agreed on caching compression type.  Any preference as to how that's implemented?

falkben:
For serving a brotli compressed precomputed volume locally and looking at it in neuroglancer, I've found that the latest version of [http-server](https://github.com/http-party/http-server) (not the outdated version on npm but the one from github) has support for brotli and will serve those files, if they exist, when a file is requested without a file extension, similar to how it serves .gz files with the `-g` flag.  For reference, this command works to serve a brotli compressed dataset locally:

`node bin/http-server PATH_TO_DATASET --cors -b -g`

william-silversmith:
Interesting! Currently, I have things set up so that you can read the .gz and .br files using `cv.viewer()`, which launches a custom web server that knows how to read those suffixes through `Storage`.

https://github.com/seung-lab/cloud-volume/blob/master/cloudvolume/server.py

william-silversmith:
I'm wary of removing the suffixes b/c the uncompressed format doesn't have a header, so by chance a given image chunk could read as a corrupted gzip or brotli file.

falkben:
> For serving a brotli compressed precomputed volume locally and looking at it in neuroglancer, I've found that the latest version of [http-server](https://github.com/http-party/http-server) (not the outdated version on npm but the one from github) has support for brotli and will serve those files, if they exist, when a file is requested without a file extension, similar to how it serves .gz files with the `-g` flag. For reference, this command works to serve a brotli compressed dataset locally:
> 
> `node bin/http-server PATH_TO_DATASET --cors -b -g`

Cool.  Another option is nice!

falkben:
> I'm wary of removing the suffixes b/c the uncompressed format doesn't have a header, so by chance a given image chunk could read as a corrupted gzip or brotli file.

Great.  Just to be clear we're only talking about local volumes, right?  None of the cloud data have extensions and we instead rely on content encoding in the headers.

william-silversmith:
> Great. Just to be clear we're only talking about local volumes, right? None of the cloud data have extensions and we instead rely on content encoding in the headers.

Yea, this only applies to the filesystem.

falkben:
Did a bit more testing -- I think this shows pretty good motivation for using brotli:

```
Size (MB)
False: 229.82
gzip: 133.16
br: 126.76

parallel = True

	Create time (s):
	False: 0.72
	gzip: 6.73
	br: 4.85

	Get time (s):
	False: 0.36
	gzip: 0.79
	br: 0.61

parallel = False

	Create time (s)
	False: 0.90
	gzip: 7.10
	br: 5.06

	Get time (s)
	False: 0.43
	gzip: 1.19
	br: 0.69
```

in summary -- seems to be faster here than gzip but giving better compression (this is a 16 bit dataset).  I need to try on an 8bit dataset.

cloudvolume/storage/storage_interfaces.py
william-silversmith: I looked into a bit more. The caching issue is tricky because not every `Storage` object is pointed at the image data. Ideally, we'd pass in a flag that enables this caching in parts of the code where such assumptions are warranted, and it would then take effect within the context of a given `get_files` call. However, this requires enough work that I'm not sure it's worth it until we experience an issue.  

Do you expect `br` to rapidly eclipse the usage of `gz`? If so, the way you've structured the if statement is ideal. Otherwise, the `.gz` test should probably go first. Based on your data below, `br` does seem advantageous.

cloudvolume/storage/storage_interfaces.py
```diff
@@ -318,6 +318,8 @@ def get_file(self, file_path, start=None, end=None):
       resp = requests.get(key, headers=headers)
     else:
       resp = requests.get(key)
+      if resp.headers.get('Content-Encoding') == 'br':
```
william-silversmith: Might be good to leave a comment here with the link to the requests and urllib3 issues.

cloudvolume/storage/storage_interfaces.py
falkben: compressing brotli sometimes takes orders of magnitude longer.  not sure it's worth it with every dataset.  hard to say if brotli will ever eclipse gzip...

william-silversmith: Is that true even at compression level 6?

falkben: I was experimenting with a uint8 dataset today and was running into some of that, yes. Should have something to share tomorrow.

william-silversmith: What about adding a way to specify the compression level?

william-silversmith:
Links to Brotli paper and experimental results. Unfortunately, most of these are on text corpus. 

https://cran.r-project.org/web/packages/brotli/vignettes/brotli-2015-09-22.pdf 
https://arpi.unipi.it/retrieve/handle/11568/938401/441027/Brotli%20paper%20%28camera%20ready%29.pdf

cloudvolume/storage/storage_interfaces.py
falkben: I made an effort today to add this as a parameter.  It's quite possible though that I may have missed some places, it needed to be added in a lot of different places!

william-silversmith: This is really cool and useful for gzip as well! :) What were your results with the uint8 data like?

falkben:
I'm a bit confused by what's going on w/ travis.

Here are some results from a test w/ the `compress_level` keyword on a small dataset (conversion to local precomputed volume from numpy array)

The essential calls were:

```python3
vol = CloudVolume(local_path, info=info, compress='br', compress_level=3, parallel=True)
start = timeit.default_timer()
data = vol[:, :, :]
stop = timeit.default_timer()
```

```
Create time (s)
('br', 1): 0.86
('br', 2): 1.06
('br', 3): 1.28
('br', 4): 3.19
('br', 5): 7.52
('br', 6): 12.89
('br', 7): 21.75
(False, None): 0.51
('gzip', None): 6.50

Size (MB)
('br', 1): 133.99
('br', 2): 132.35
('br', 3): 132.25
('br', 4): 132.23
('br', 5): 127.00
('br', 6): 126.25
('br', 7): 126.22
(False, None): 229.82
('gzip', None): 136.68```

File: cloudvolume/storage/storage_interfaces.py (hunk 1 above)
Comment (falkben) [reply cloudvolume/storage/storage_interfaces.py]: Compression rate good, decompression rate good, but for some reason compression time went way up for levels > 4.  Investigating further on another machine.

Comment (falkben) [reply cloudvolume/storage/storage_interfaces.py]: Brotli compression is single threaded as far as I know.  I'm specifying cloudvolume w/ Parallel=False.  Still seeing lots of cores being used.

On two machines, seeing a big difference in performance w/ brotli, but seemingly only with the 8-bit data.

Below are the two different runs:

32 thread (16 cores) server:
```
Create time (s)
('br', 1): 8.20
('br', 2): 7.41
('br', 3): 7.74
('br', 4): 7.77
('br', 5): 8.04
('br', 6): 9.18
('br', 7): 13.71
(False, None): 6.61
('gzip', None): 7.59

Size (MB)
('br', 1): 444.20
('br', 2): 425.68
('br', 3): 425.15
('br', 4): 436.03
('br', 5): 349.76
('br', 6): 347.46
('br', 7): 347.62
(False, None): 733.61
('gzip', None): 392.54
```

8 threads (4 cores) laptop:
```
Create time (s)
('br', 1): 4.27
('br', 2): 4.23
('br', 3): 4.25
('br', 4): 5.60
('br', 5): 23.29
('br', 6): 43.60
('br', 7): 66.36
(False, None): 3.23
('gzip', None): 7.10

Size (MB)
('br', 1): 444.20
('br', 2): 425.68
('br', 3): 425.15
('br', 4): 436.03
('br', 5): 349.76
('br', 6): 347.46
('br', 7): 347.62
(False, None): 733.61
('gzip', None): 392.54
```

The compression ratio is at least the same across the two different machines.

Comment (falkben) [reply cloudvolume/storage/storage_interfaces.py]: Further tests make be believe that there's nothing specific about 8 bit data that is a problem -- perhaps it's just that a larger dataset and the number of threads available that slows down the compression.

Here's some tests on the exact same 16 bit dataset, both at 16 bit and also coerced to 8 bit.  

In this case, I do not see anything peculiar about brotli with 8 bit data, and it actually performs much faster.  I think changing brotli default quality=5 might be better though.  It will be closer in performance to gzip I think, but still give better compression.

`16bit`
```
Create time (s)
('br', 1): 0.81
('br', 2): 0.93
('br', 3): 1.09
('br', 4): 2.95
('br', 5): 7.21
('br', 6): 12.69
('br', 7): 21.45
(False, None): 0.57
('gzip', None): 6.30

Size (MB)
('br', 1): 133.99
('br', 2): 132.35
('br', 3): 132.25
('br', 4): 132.23
('br', 5): 127.00
('br', 6): 126.25
('br', 7): 126.22
(False, None): 229.82
('gzip', None): 136.68
```

`8bit`
```
Create time (s)
('br', 1): 0.64
('br', 2): 0.65
('br', 3): 0.70
('br', 4): 0.89
('br', 5): 1.49
('br', 6): 2.06
('br', 7): 2.43
(False, None): 0.53
('gzip', None): 1.32

Size (MB)
('br', 1): 93.38
('br', 2): 93.36
('br', 3): 93.20
('br', 4): 92.95
('br', 5): 93.10
('br', 6): 93.11
('br', 7): 93.10
(False, None): 114.91
('gzip', None): 93.61
```

Comment (falkben) [reply cloudvolume/storage/storage_interfaces.py]: I don't really know what to make of this, however.  Seems like cloud-volume uses multiple cores (no matter what setting is specified) and that can make compression benchmarking w/ cloudvolume really difficult.

16 cores, 32 threads

`16bit`
```
Create time (s)
('br', 1): 1.44
('br', 2): 1.37
('br', 3): 1.45
('br', 4): 1.64
('br', 5): 2.16
('br', 6): 3.26
('br', 7): 5.69
(False, None): 1.25
('gzip', None): 3.71
```

`8bit`
```
Create time (s)
('br', 1): 1.26
('br', 2): 1.17
('br', 3): 1.21
('br', 4): 1.21
('br', 5): 1.30
('br', 6): 1.40
('br', 7): 1.46
(False, None): 1.01
('gzip', None): 1.46
```

**Comment by william-silversmith** [comment]
The same error is present in `test_compression`

**Comment by falkben** [comment]
> 
> 
> ~The same error is present in `test_compression`~ Oh no. It's actually that Google Cloud Python isn't decoding `br` encoded files despite this issue:
> 
> [googleapis/google-cloud-python#9003](https://github.com/googleapis/google-cloud-python/issues/9003)
> 
> `google-resumable-media` is pegged in this repo, but I tested with upgrading to 0.4.1... still broken.

Disable brotli for now in the google storage interface?

**Comment by william-silversmith** [comment]
That's the fastest way to get this working. Just disabling Google will enable this for s3, matrix, and file which should get things working for you. My lab mostly uses GCS, but given you're the patch contributor we should prioritize your needs. In a separate PR, I might consider writing a patch for google-cloud-python and use the fork. We can contribute it upstream and see if they like it.

**Comment by william-silversmith** [comment]
@falkben looks like Google will be implementing a way to get the content encoding in the near future. Let's proceed with blocking GCS for now and implement the functionality when it's available.

File: cloudvolume/storage/storage_interfaces.py [hunk 3]
```diff
@@ -190,10 +190,8 @@ def put_file(self, file_path, content, content_type, compress, cache_control=Non
     key = self.get_path_to_file(file_path)
     blob = self._bucket.blob( key )
 
-    # keep default gzip
-    if compress == "br":
-      blob.content_encoding = "br"
-    elif compress:
+    # gcloud disable brotli until content-encoding works
```
william-silversmith: I don't think this is the right patch b/c (a) we only have a problem with decompression, not compression (b) if `br` is selected for one file type this will transparently change to `gzip` without notifying the programmer something weird happened. I was thinking more along the lines of, perhaps we should raise an exception in CloudVolume if `compress='br'` for example. What do you think?

falkben: yes, much better idea.

falkben:
Seemed like there's nothing to be done about getting `br` compressed objects on GCS (there shouldn't be any anyways, right?)

cloudvolume/storage/storage_interfaces.py
```diff
@@ -191,7 +191,9 @@ def put_file(self, file_path, content, content_type, compress, cache_control=Non
     blob = self._bucket.blob( key )
 
     # gcloud disable brotli until content-encoding works
-    if compress:
+    if compress == "br":
+      raise TypeError("Brotli unsupported on google cloud storage")
```
william-silversmith: Sorry to be overly pedantic, but I think this is probably more of a `ValueError` than a `TypeError`. However, it would be better to create a new exception class, something like `exceptions.UnsupportedCompressionType` in `exceptions.py` which is more self-explanatory in code that tests for it in a try-catch and also more specific.

william-silversmith:
Yea, until we have the `Content-Encoding` we can't do anything about decompression without some ugly workarounds.

cloudvolume/storage/storage_interfaces.py
falkben: makes sense, was in a meeting when i coded that up. fixed

william-silversmith:
Thanks for making those changes! I gotta admit, the triple checks for the `file://` protocol are a bit disturbing to me. If this keeps up I'll have to figure out a new file format or something (maybe a single file extension with a few bytes header?). I can implement it once someone complains. ^_^

Thanks for all your hard work shepherding this through @falkben. It's gonna be lit. :)

PR #287, opened 2019-11-25
fix: missing import for http interface

falkben:
This PR fixes a missing import for brotli support on the http interface.

Also adds a test depending on neurodata open data bucket (needed to be a brotli dataset).


william-silversmith:
Version 0.65.5 contains this update.

PR #289, opened 2019-11-26
fix: accurate mip level and voxel offset accomodation for vertex deduplication

william-silversmith:
We can also extract mip from the mesh path since Igneous has
been using the directory name to store the mip level used for
meshing.

e.g. mesh_mip_2_err_40

err_40 refers to allowing up to 40nm of displacement during
simplificiation.


william-silversmith:
Discussed the draco changes with @manuel-castro and it seems we need to provide the draco chunking for the appropriate graph level from the PyChunkGraph server.

william-silversmith:
I'm going to move the Draco part of this PR to an Issue and we'll address it in another PR https://github.com/seung-lab/cloud-volume/issues/299

fcollman:
I'd like to see the tests from #297 incorporated first before this.

PR #291, opened 2019-11-27
install: requests/urllib supports brotli

falkben:
Somehow I missed that `urllib3/requests` does now support `brotli`, through the optional install parameter in urllib3 (and by using latest releases from both).  Discovered this yesterday when I had a different version of requests installed in a different virtual environment and was running into problem decoding brotli encoded data (because it was trying to decode an already decoded response, and I had brotli installed in that environment).

A few things to keep an eye out for w/ this PR:

1. This "unpins" and upgrades `urlib3`.  Also upgrades requests, but that wasn't unpinned, so don't think it's a big deal.  Not sure if pinning urlib3 was important, can repin if necessary.  Brotli was added in 1.25 of urlib3.  I tested on 1.25.7.  Requests added support in latest release, 2.22.0.

2. I'm not sure what this does for people who install from source.  I think it will be fine though, as long as they reinstall from requirements.txt

3. I didn't/can't run the entire test suite

### DIFF
--- a/cloudvolume/storage/storage_interfaces.py
+++ b/cloudvolume/storage/storage_interfaces.py
@@ -12,7 +12,6 @@ import google.cloud.exceptions
 from google.cloud.storage import Batch, Client
 import requests
 import tenacity
-import brotli
 
 from cloudvolume.connectionpools import S3ConnectionPool, GCloudBucketPool
 from cloudvolume.lib import mkdir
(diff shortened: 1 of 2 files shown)


falkben:
This will hopefully be the last brotli PR!

william-silversmith:
Thanks for the upgrades!

> This will hopefully be the last brotli PR!

I'm still waiting on google-cloud-python to return the Content-Encoding with a request before adding brotli support. ;)

william-silversmith:
Tests pass on my machine for python3.7.

william-silversmith:
`import brotli` will be making a comeback for GCS support in all probability.

PR #292, opened 2019-11-28
fix: EmptyVolumeException w/ non_aligned_writes

### LINKED ISSUE #290: EmptyVolumeException during non-aligned writes
I have input tiles that overlap, so I am trying to do non-aligned writes rather than create aligned tiles myself.   I get an EmptyVolumeException in the download code, which is unaffected by the fill_missing value.  The code below reproduces the problem -- the final statement throws the exception.

I have played around with the size of the data and tiles, and the exception doesn't always get thrown on the 2nd tile write -- but in practice I've seen it thrown on some tile during my upload.

```
import cloudvolume
import numpy as np

path = 'file:///media/ssd/data/testing/empty_volume_exception'

data_extent = 2048
stride = 2028

shape = (data_extent, data_extent, 1, 1)
data = np.zeros(shape, dtype=np.uint16)

full_shape = (stride + data_extent, stride + data_extent, 1, 1)

info = cloudvolume.CloudVolume.create_new_info(
    num_channels=data.shape[3],
    layer_type='image',
    data_type=data.dtype.name,
    encoding='raw',
    chunk_size=(256, 256, 32),
    volume_size=full_shape[:3],
    max_mip=0,
    resolution=[1,1,1],
    voxel_offset=[0,0,0],
)
vol = cloudvolume.CloudVolume(path, info=info, parallel=1, progress=False, 
                              non_aligned_writes=True, fill_missing=True)
vol[slice(0, data_extent), slice(0, data_extent), :, :] = data
vol[slice(stride, stride+data_extent), slice(0, data_extent), :, :] = data
```

william-silversmith:
Resolves #290

### DIFF
--- a/cloudvolume/datasource/precomputed/image/__init__.py
+++ b/cloudvolume/datasource/precomputed/image/__init__.py
@@ -176,6 +176,7 @@ class PrecomputedImageSource(ImageSourceInterface):
       background_color=self.background_color,
       non_aligned_writes=self.non_aligned_writes,
       green=self.config.green,
+      fill_missing=self.fill_missing, # applies only to unaligned writes
     )
 
   def exists(self, bbox, mip=None):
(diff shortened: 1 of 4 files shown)


william-silversmith:
Been thinking about this and I'm pretty sure the `fill_missing=False` was done out of laziness rather than a well grounded reason. I can't think of a reason that a missing file would mean something different for non-aligned writes than it would for a normal upload.

PR #297, opened 2019-12-02
fix+test: Graphene mesh tests & fixes subclassing

fcollman:
This adds tests for graphene that uses actual mesh fragments.  I have added draco compressed versions of fragments as well for the same mesh case that involves a self contact.  Opening the PR now for overall feedback.. this also appears to fix this test at least for graphene, though I don't think this is the correct overall architecture.. so happy to take feedback or have someone else build off of this test to make an alternative fix work.


test/test_graphene.py
```diff
@@ -5,11 +5,84 @@
...
 import pytest
+import os
+from scipy import sparse 
 
 tempdir = tempfile.mkdtemp()
 TEST_PATH = "file:/{}".format(tempdir)
 TEST_DATASET_NAME = "testvol"
+MESH_TEST_DATASET_NAME = "meshvol"
 PCG_LOCATION = "http://localhost/segmentation/1.0/"
+PCG_MESH_LOCATION = "http://localhost./meshing/1.0/"
```
william-silversmith: localhost `.` /meshing?

fcollman: i have no idea about this.. when i took this off it complained that my mock wasn't formatted correctly...

fcollman: I figured it out.. was how the meshing url was constructed... fixed now

test/test_cv/info
```diff
@@ -0,0 +1,202 @@
+{
+  "data_dir": "file:///Users/forrestc/ConnectomeStack/cloud-volume/test/test_files/",
```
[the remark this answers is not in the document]
fcollman: yeah i think the present code doesn't even use this info file, i just thought it was useful for testing but i obviously can't anticipate where the user will check their code out..

test/test_graphene.py
```diff
@@ -5,11 +5,81 @@
...
+        "voxel_offset": [
+            17920,
+            14848,
+            0
+        ]
+        }],
+        "type": "segmentation"
+    }
+    requests_mock.get(PCG_LOCATION+MESH_TEST_DATASET_NAME+"/info", json=info_d)
+    requests_mock.get(PCG_LOCATION+MESH_TEST_DATASET_NAME+"/info/", json=info_d)
```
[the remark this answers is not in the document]
fcollman: I think this is just to cover both instances of mock its not strictly necessary

william-silversmith:
What python version is giving you the deprecation warning? I don't see it in py36

william-silversmith:
Looks like both spellings are available since python2.6, so this change is quite safe.

https://docs.python.org/2/library/threading.html#threading.Thread.is_alive

william-silversmith:
The `is_alive` change was moved to #300

Issue #264, opened 2019-08-30
Improve Performance of Sharded Reads


william-silversmith:
Uncached sharded reading is several times slower than standard reads (e.g. 3-5 per sec vs 20+ per sec). This state of affairs can be improved by a few methods:

- [ ] Use an in-memory LRU cache for the shard index
- [ ] Use threads / reactive programming to pipeline index and data fetching
- [ ] Condense requests targeting the same shard into fewer longer reads if they are nearby.

william-silversmith:
#347 is related.

PR #298, opened 2019-12-02
perf: cache shard and minishard indices

william-silversmith:
With default settings assuming 9 bits for each in the spec:
Shard index cache will take up about 4MB (512 entries)

Minishard index cache will take up potentially more... (128 entries) It is a bit less useful than the shard index cache, but for example, I ran across a wild example of a spec with 0 shard bits and many repeated accesses to the same minishards. A small cache will speed up odd cases like this.

Introduces new dependency on `lru-dict`

Partially addresses #264


william-silversmith:
Should replace the C-extension `lru-dict` with `pylru` to ensure it works cross platform since there are no binaries deployed for `lru-dict`.

PR #300, opened 2019-12-03
changing spelling of isAlive>is_alive

fcollman:
should address a deprecation warning


william-silversmith:
This is supported since python 2.6.

PR #308, opened 2019-12-20
Create methods to convert to/from dask arrays.

chrisroat:
test_roundtrip_4d_channel_rechunked is failing


chrisroat:
Theses functions have specific opinions on how to handle chunking and shaping of arrays, particularly in regard to the channel dimension.  Please feel free to make suggestions if you think it should be handled differently.

Tests are included, though one dealing with channel chunking is still failing.  I could use some help digging into it.

I wasn't sure where you'd want these methods, so I included them in a separate file.   I'm happy to move them elsewhere.

cloudvolume/dasklib.py
```diff
@@ -0,0 +1,133 @@
...
+    vol.provenance.processing = [{'method': 'dask_to_cloudvolume'}]
+    vol.commit_provenance()
+    return arr.store(vol,
+                     lock=False,
+                     compute=compute,
+                     return_stored=return_stored)
+
+
+def dask_from_cloudvolume(vol_path, chunks=None, name=None, **kwargs):
+    """Load dask array from the cloudvolume storage format.
```
william-silversmith: CloudVolume supports several storage formats. It might be more accurate to say from a cloudvolume compatible dataset.

cloudvolume/dasklib.py
```diff
@@ -0,0 +1,133 @@
...
+    vol.commit_info()
+    vol.provenance.processing = [{'method': 'dask_to_cloudvolume'}]
+    vol.commit_provenance()
+    return arr.store(vol,
+                     lock=False,
+                     compute=compute,
+                     return_stored=return_stored)
+
+
+def dask_from_cloudvolume(vol_path, chunks=None, name=None, **kwargs):
```
william-silversmith: In other areas of the codebase, we tend to call `vol_path` `cloudpath`. It would be good to standardize this (though the from_numpy function might diverge...).

test/test_dasklib.py
```diff
@@ -0,0 +1,93 @@
+import pytest
+
+import numpy as np
+
+from cloudvolume import dasklib
```
william-silversmith: This could be `import cloudvolume.dask as dasklib`

chrisroat:
Thanks for the review.  I've addressed most of the changes.

The one change I wasn't sure about was adding `to_dask` to the `CloudVolumePrecomputed` class.  In particular, what were you thinking would be more optimal, as compared to the `cloudvolume.dask.to_dask` method that has been added?

The `test_roundtrip_4d_channel_rechunked` test is still failing, and I haven't yet dug into it.  If your experience with cloudvolume gives you a clue as to the problem, do let me know.

chrisroat:
To answer your question: the workloads O(TB) volumetric confocal images.  I'm only doing writing at the moment, as cloud-volume is my "viewer" format and I will continue to use zarr for computation read/write.  From a 16-core GCE instance, I can write 10GB to zarr N5 on GCS in 30 seconds, while cloud-volume (using this PR) on GCS takes about 15 minutes.

Cloud-volume may be optimized for doing it's own parallelization, and perhaps something isn't ideal when dask tasks are writing -- and it's possible I've overlooked something here.

william-silversmith:
Thanks Chris! I'm on vacation now but I'll look when I get back.

It sounds like CloudVolume is using one core vs 16 in use by zarr assuming
the same compression method. In my work, either I use parallel=~number of
cores or we use Kubernetes to associate e.g. 16 single core tasks to the
instance so all cores are fully utilized.

GCE virtualized instances are also very bad for CloudVolume's threading
when there are many more idle cores than utilized as it causes excessive
thrashing. The scheduler behaves more nicely when each process's threads
are forcibly confined to a single core. CV green threads fix this problem
as does the taskset system command which can pin a process to a core at
parallel 1. This issue is not present on my local non-virtualized machine.

Dask is a slightly different computing model and I'll have to review.

Happy New Year!

chrisroat:
Happy New Year!

Dask is using multiple threads, and the code uses the default of parallel=1 for each thread.    I can confirm this by looking at the dask profiles.  One change is to see if using one process per worker (rather than thread) helps.   That could provide a hint, though we should still figure out how to make the thread-based setup work.

Perhaps there something else that is causing the slowdown.  I believe dask should just be calling something like `cv[slices] = data_chunk` in each task (thread or process).

william-silversmith:
I think if you are using threads, you'll run into the cPython Global Interpreter Lock (GIL) which will mean only one thread can run at a time. For Dask computations, it probably mostly defaults to numpy code (or other C extension code) that releases the GIL and enables real threaded computation. Since CloudVolume is mostly Python code that runs 20 threads itself per a process, you'll only be able to process a small amount at a time.

Using multiple processes is the way to go.

chrisroat:
I can verify your hypothesis on a machine  with 18 core / 36 hyperthread, just using file output.  Using single-threaded cloudvolume with the dask `threads` scheduler is quite slow.  A 8x8x8 volume with 64 chunks takes 3 seconds to save on my machine!!  Switching to the `processes` scheduler drops that 0.25s, which is mostly overhead.   A 512x512x512 volume, again with 64 chunks, is 5s (`threads`) and 0.5s (`processes`).

```
import dask
import dask.array as da
from dask.diagnostics import Profiler, ResourceProfiler, CacheProfiler, ProgressBar, visualize
import bokeh.io
import cloudvolume.dask as dasklib

bokeh.io.output_notebook()

data = da.zeros((512,512,512), chunks=128)  # 64 chunks
tasks = dasklib.from_dask(data, 'file:///media/ssd/data/dask_cv/test1', compute=False)

with Profiler() as prof, ResourceProfiler(dt=0.25) as rprof, CacheProfiler() as cprof, ProgressBar():
    dask.compute(tasks, scheduler='threads')  # <--scheduler

visualize([prof, rprof, cprof])
```

william-silversmith:
Thanks for testing. I should probably write a wiki article on the purpose of green threads in the CloudVolume ecosystem. I'm looking more closely at the interface right now and there's a few points that come to mind.

1) in `from_dask`, is it the case that the `compute` and `return_stored` are mutually exclusive? Why not have a single flag that flips the return type? 
2) `from_dask` behaves a bit differently than I expect, but it makes sense for the application as Dask will continue to be used from that point on. Ordinarily, I would have expected it to return a CloudVolume instance. e.g. `cv = from_dask(arr)`
3) The reason I would have though a `to_dask` method would be nice would be so that you could get a dask array trivially from a CloudVolume instance like so: `arr = cv.to_dask()`
4) The return type is missing from the documentation for `to_dask` 
5) I think what is happening with the error on `test_roundtrip_4d_channel_rechunked` is that CloudVolume treats num_channels differently from chunks (this was the design from neuroglancer). CloudVolume will not chunk along the fourth axis, each file contains all channels specified. If you set `arr = da.zeros((3, 3, 3, 3), chunks=(2,2,2,3))` the test works.

chrisroat:
1. The interface comes from dask itself - the two flags are independent.  [This code]( https://github.com/dask/dask/blob/27f5bed5f7f8c03a5ef506a408cff2db8e9ca26b/dask/array/core.py#L902) shows how the two flags interact.  `return_stored` determines what the return value is.  `compute` determines if the computation is performed immediately or not.  For example, if you specify `compute=True` and `return_stored=False`, it returns None.
1. I updated the names to reflect better what they do and to fit the dask ecosystem.  
1. I totally agree.  Added.  Your previous comments phrasing made me think there was something special about that object that would make it more efficient.  
1. Added return sections in both functions
1. Updated the test.  Thanks for the explanation.

william-silversmith:
These changes make sense to me! Does it make sense to ever set `return_stored=False` then? It seems kind of useless to return None.

william-silversmith:
Oh, I see. If you set `return_stored`, Dask will attempt to merge the results from several machines, so there is a performance hit.

william-silversmith:
Thanks for your hard work on this Chris!

PR #309, opened 2020-01-17
Delay creating metadata in dask.to_cloudvolume 

chrisroat:
I was finding a huge slowdown when opening many volumes for writing on a cloud filesystem.  All the metadata files were being written from my machine -- instead of being sharded to workers.  

In addition, when compute=False, it may not be the intention of the caller to ever write out data.  The PR delays the writing of `info` or `provenance` until compute time.

For these reasons, I think we can delay writing the metadata.  It's an open question to me if we should always wrap in `dask.delayed`, regardless of the value of compute.  I believe this may be better, and am trying to open a discussion with the dask maintainers on a similar PR in their repo - dask/dask#5797

### DIFF
--- a/cloudvolume/dask.py
+++ b/cloudvolume/dask.py
@@ -63,6 +63,7 @@ def to_cloudvolume(arr,
   -------
   See notes on `compute` and `return_stored` parameters.
   """
+  import dask
   import dask.array as da
   if not da.core._check_regular_chunks(arr.chunks):
     raise ValueError('Attempt to save array to cloudvolume with irregular '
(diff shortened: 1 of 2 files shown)


chrisroat:
I've updated this PR to always "delay" the initial creation of info and provenance metadata.  In reality, all this does is create a dependency such that the metadata is written prior to the data.  It simply moves metadata creation to compute-time instead of during dask graph-computation time.

For immediate computation, this shouldn't show an effect.  For delayed computation, this will move the creation to a worker.

I think it is ready to go.  The travis failure seems unrelated.

william-silversmith:
lgtm, tests pass when running with the appropriate credentials

PR #318, opened 2020-03-20
feat: making sharded mesh readers for graphene, added decoder for labels

william-silversmith:
We are adding the capability to decode graphene segids and read "initial" L2 meshes that will be created in the sharded format.


william-silversmith:
Plotting two different fan outs assuming # shards = # of L2 chunks. X axis is "levels" and Y axis is "shard bits"

image

william-silversmith:
Wrote a more careful analysis here: https://github.com/seung-lab/cloud-volume/wiki/Graphene#the-sharded-format

Issue #280, opened 2019-10-22
Why is resolution int when Neuroglancer supports floats?


falkben:
Just wondering if there was a reason resolution gets coerced into an `int` here? 

https://github.com/seung-lab/cloud-volume/blob/master/cloudvolume/datasource/precomputed/metadata.py#L112

We have datasets with non integer resolution values and Neuroglancer precomputed supports floats for the scale keys so just curious if there was a reason for this before I submitted PR.  I suppose actual int keys would become `x.0` which is not as clean if it was a straight replacement to `float`

william-silversmith:
At the time this was written (2017), we weren't sure if neuroglancer supported floating point resolution. It may have, but we thought that our use cases probably wouldn't need it given we were talking about single digit nanometers. I am currently working with a dataset that would prefer fractional units, so it's possible this may change. There may be a good deal of software written with integer units in mind though.

william-silversmith:
I am currently working on this, but it is a slow process to make sure everything is working in both CV and Igneous before releasing it.

william-silversmith:
This was resolved in https://github.com/seung-lab/cloud-volume/pull/467!

PR #322, opened 2020-04-10
Support for non-int resolutions

AF1111:
Hi! I noticed #280  was never addressed with a pr and was wondering if this change is something you would want? I'm also working with float resolution volumes where the precision matters for matching up fragments.

### DIFF
--- a/cloudvolume/datasource/precomputed/metadata.py
+++ b/cloudvolume/datasource/precomputed/metadata.py
@@ -109,7 +109,7 @@ class PrecomputedMetadata(object):
         "encoding": encoding,
         "chunk_sizes": [chunk_size],
         "key": "_".join(map(str, resolution)),
-        "resolution": list(map(float, resolution)),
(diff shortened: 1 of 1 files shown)


william-silversmith:
Hi Alex, this is a good PR. I think it will require some testing with Igneous to make sure that this doesn't break anything, but I think my gut take is we can probably merge this. I'll get back to you in a few days when I have a chance to do some testing.

william-silversmith:
https://github.com/seung-lab/igneous/issues/55 

Looks like there is an issue in Igneous that would need to be fixed.

william-silversmith:
It seems that even if Igneous doesn't support it, people are clamoring for non-integer resolutions. I think there are some fixes that are required to make this backwards compatible so I'm going to close this issue and reuse some of the contributed code on a new PR that I can edit. I'll give you a credit Alex in the README for helping towards integrating this feature.

william-silversmith:
(and Ben too!)

PR #323, opened 2020-04-10
Add range read request to graphene and new properties to graphene metadata

manuel-castro:
Range read request is needed to do meshing

Edit: This PR is reliant on this PR in the CG to be deployed: https://github.com/seung-lab/PyChunkedGraph/pull/198

### DIFF
--- a/cloudvolume/frontends/graphene.py
+++ b/cloudvolume/frontends/graphene.py
@@ -292,7 +292,7 @@ class CloudVolumeGraphene(CloudVolumePrecomputed):
 
     version = GrapheneApiVersion('v1')
     path = version.path(self.meta.server_path)
-    url = posixpath.join(self.meta.base_path, path, "range_read_binary", str(chunk_id))
+    url = posixpath.join(self.meta.base_path, path, "l2_chunk_children_binary", str(chunk_id))
 
     response = requests.get(url, params={}, headers=self.meta.auth_header)
     response.raise_for_status()
--- a/cloudvolume/frontends/graphene.py
+++ b/cloudvolume/frontends/graphene.py
@@ -284,11 +284,11 @@ class CloudVolumeGraphene(CloudVolumePrecomputed):
 
     layer_id = self.meta.decode_layer_id(chunk_id)
     if layer_id != 2:
-      raise ValueError("This function only accepts layer 2 chunk IDs. Got {}".format(self.meta.decode_label(chunk_id)))
+      raise ValueError("This function only accepts Layer 2 chunk IDs. Got {}".format(self.meta.decode_label(chunk_id)))
 
     segid = self.meta.decode_segid(chunk_id)
     if segid != 0:
-      raise ValueError("Chunk IDs have zeroed segid. Got: {}".format(self.meta.decode_label(segid)))
+      raise ValueError("Chunk IDs have a zeroed segid field. Got: {}".format(self.meta.decode_label(segid)))
 
     version = GrapheneApiVersion('v1')
     path = version.path(self.meta.server_path)
--- a/cloudvolume/frontends/graphene.py
+++ b/cloudvolume/frontends/graphene.py
@@ -264,7 +264,7 @@ class CloudVolumeGraphene(CloudVolumePrecomputed):
 
     return np.array(roots, dtype=self.meta.dtype)
 
-  def get_range_read(self, chunk_id, timestamp=None):
+  def get_seg_ids_in_chunk(self, chunk_id, timestamp=None):
     if isinstance(timestamp, str):
       timestamp = dateutil.parser.parse(timestamp) # returns datetime
     if isinstance(timestamp, datetime): # NB. do not change to elif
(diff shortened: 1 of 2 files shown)


william-silversmith:
Range read means segids? In the context of sharding it means byte ranges. Maybe we should name this a bit differently to avoid the conflict and be less ambiguous.

william-silversmith:
There's a little bit of conflict with #318, we'll have to merge somehow.

william-silversmith:
Can you also add a docstring to the new function?

manuel-castro:
> Range read means segids? In the context of sharding it means byte ranges. Maybe we should name this a bit differently to avoid the conflict and be less ambiguous.

Yes, in the context of CG it means seg ids. But I'll change it to say `read_chunk_seg_ids` here, I agree that that's more intuitive.

william-silversmith:
I made a few changes in order to rebase on master, let me know what you think!

manuel-castro:
> I made a few changes in order to rebase on master, let me know what you think!

Looks good! I added a docstring and changed the function name.

william-silversmith:
Interesting! I didn't realize this was for L2 only. Is there something analogous for higher mip levels?

manuel-castro:
In the CG you can use it for higher mip levels too, but we are limiting the CG endpoint to lvl 2 id requests to lessen the load on the CG (and only lvl 2 are needed to do meshing)

cloudvolume/frontends/graphene.py
```diff
@@ -261,6 +262,33 @@ def get_roots(self, segids, timestamp=None, binary=True, stop_layer=None):
 
     return np.array(roots, dtype=self.meta.dtype)
 
+  def get_seg_ids_in_chunk(self, chunk_id, timestamp=None):
+    """
+    Get all of the segments in a lvl 2 chunk,
+    as well as their associated supervoxels.
+
+    chunk_id: np.uint64 lvl 2 chunk id
```
william-silversmith, line 270: I think we need to be careful to describe what a chunk_id is. In the rest of this section, a chunk_id is defined as given a label <LAYER,X,Y,Z,SEGID>, the chunk_id is uint64(X << 20 |Y << 10 | Z) (assuming 10 is the offset for this graph). I noticed in some other cost samples you gave me, you defined the chunkid as meta.encode_label(LAYER,X,Y,Z,segid=0). I'm assuming here we are using the latter? Should we reconcile these definitions?

william-silversmith: Talked about this with Manuel. Chunk ID is defined as LAYER/X/Y/Z/0 (ie segid zeroed out). I'll call the chunk_id I was using `chunk_position_number`

william-silversmith:
Useful L2 label: 173729460028182040

william-silversmith:
Failed test is spurious.

william-silversmith:
@manuel-castro do you think `get_seg_ids_in_chunk` should accept any label or only proper chunk IDs? We can generate a proper chunk ID in the function.

william-silversmith:
Also thinking of changing the name to `get_chunk_labels` just because its shorter.

william-silversmith:
It looks like the pickled response requires `google-cloud-bigtable` to be installed. Is that necessary?

william-silversmith:
Can you help me interpret the output?

william-silversmith:
I merged in the metadata changes as I'll need them sooner.

william-silversmith:
I added more documentation and renamed it again from `get_l2_chunk_mappings` to `get_chunk_mappings` since the latter is more easily extensible and we already check for L2 chunk ids by raising an exception.

cloudvolume/frontends/graphene.py
```diff
@@ -263,6 +277,49 @@ def get_roots(self, segids, timestamp=None, binary=True, stop_layer=None):
...
+    chunk_id = self.meta.decode_chunk_id(chunk_id)
+    
+    version = GrapheneApiVersion('v1')
+    path = version.path(self.meta.server_path)
+    url = posixpath.join(self.meta.base_path, path, "l2_chunk_children_binary", str(chunk_id))
+
+    response = requests.get(url, params={}, headers=self.meta.auth_header)
+    response.raise_for_status()
+
+    return pickle.loads(response.content)
```
william-silversmith, line 321: I'm still very leery of this `pickle.loads` in client-side code. I would much rather a more well defined parser.... If someone captured or redirected DNS to the graph server, they could achieve remote code execution using this.

william-silversmith: What about http://bsonspec.org/faq.html ? I tried an example from this function and got: 
```
710 kB with JSON
125 kB JSON gzipped
490 kB with BSON
148 kB with BSON gzipped
```

william-silversmith: Ofc, it might be even more efficient to return { watershed_id: l2_id } and then encode it in a simple array as [ ws1, l2_1, ws2, l2_2, ... ]

manuel-castro: > Ofc, it might be even more efficient to return { watershed_id: l2_id } and then encode it in a simple array as [ ws1, l2_1, ws2, l2_2, ... ]

This is what I ended up doing. Rather [l2_1, ws1_1, l2_1, ws1_2, ... ]

william-silversmith: That's good but that reversal will inflate the size of the binary substantially. I'm OK with that for the sake of practicality, but it might defeat the point binary.

william-silversmith: Never mind, I confused myself.

PR #324, opened 2020-04-15
feat: add image write locking per mip level 

william-silversmith:
This sets the property "locked" to true in each selected mip within info['scales']. It is then checked when calling upload or delete.

This mechanism is very vulnerable to race conditions, and is mainly intended for an additional safety mechanism to prevent accidental deletion. It's not meant to be called frequently.

### DIFF
--- a/cloudvolume/datasource/precomputed/metadata.py
+++ b/cloudvolume/datasource/precomputed/metadata.py
@@ -703,7 +703,7 @@ Hops:
       for mip in mips:
         self.info['scales'][mip]['locked'] = False
 
-      self.commit_info(info)
+      self.commit_info()
     except Exception as err:
       msg = lib.yellow("Unable to release lock on mips {}".format(list(mips)))
       raise exceptions.WriteLockReleaseError(msg)
--- a/cloudvolume/datasource/precomputed/metadata.py
+++ b/cloudvolume/datasource/precomputed/metadata.py
@@ -709,4 +709,4 @@ Hops:
       raise exceptions.WriteLockReleaseError(msg) from err
 
   def locked_mips(self):
-    return set([ i for i, scale in enumerate(self.info['scale']) if scale.get('locked', False) ])
+    return set([ i for i, scale in enumerate(self.info['scales']) if scale.get('locked', False) ])
--- a/cloudvolume/datasource/precomputed/metadata.py
+++ b/cloudvolume/datasource/precomputed/metadata.py
@@ -146,10 +146,14 @@ class PrecomputedMetadata(object):
     
     return info
 
-  def refresh_info(self, max_redirects=10):
+  def refresh_info(self, max_redirects=10, force_fetch=False):
     """
     Refresh the current info file from the cache (if enabled) 
     or primary storage (e.g. the cloud) if not cached.
+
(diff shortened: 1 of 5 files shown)


william-silversmith:
Test Plan:

- Lock a mip level in a test volume and ensure neuroglancer still displays it
- Try to write to it, and make sure it raises ReadOnlyException
- Unlock that mip level it and make sure it can display and write
- Lock another mip level and make sure it can display and write to the previous mip level

cloudvolume/datasource/precomputed/metadata.py
```diff
@@ -658,4 +662,51 @@ def add_scale(self, factor, encoding=None, chunk_size=None, info=None):
...
+        max(self.available_mips), mips
+      ))
+
+    try:
+      self.refresh_info(force_fetch=True)
+
+      for mip in mips:
+        self.info['scales'][mip]['locked'] = False
+
+      self.commit_info(info)
```
nkemnitz, line 706: `info` is not defined.

cloudvolume/exceptions.py
```diff
@@ -124,4 +124,12 @@ class SpecViolation(Exception):
 
 class SpatialIndexGapError(Exception):
   """Part of the spatial index was not found. A complete result set cannot be fetched."""
+  pass
+
+class WriteLockAcquisitionError(Exception):
```
nkemnitz, line 129: You might want to think about inheritance for your exceptions, so that it's easier to catch groups of exceptions

cloudvolume/datasource/precomputed/metadata.py
```diff
@@ -658,4 +662,51 @@ def add_scale(self, factor, encoding=None, chunk_size=None, info=None):
...
+    return newscale
+
+  def lock_mips(self, mips):
+    """
+    Establishes a write lock on the specified mip levels.
+    The lock is written to the cloud info file.
+    """
+    mips = lib.toiter(mips)
+    if max(mips) > max(self.available_mips):
+      raise ValueError("Cannot lock a mip level that doesn't exist. Highest mip: {} Got: {}".format(
```
nkemnitz, line 674: Maybe I want to prevent the generation of additional downsamples? 😈

cloudvolume/exceptions.py
william-silversmith: This is a good idea. I'm not sure what the right hierarchy is right now though. It would be helpful to have a use case.

cloudvolume/datasource/precomputed/metadata.py
william-silversmith: Is that realistic? I'm not sure why you would care as you could delete them.

william-silversmith:
I guess for SEAMLeSS and its ilk, we should be more careful to add checks for this. For certain datasets, we might want to use bucket level locking like you recommended earlier. The nice thing about this approach though, is that it works regardless of storage system, including standard FS.

william-silversmith:
Neuroglancer seems like it's cool with the additional locked parameter.

cloudvolume/datasource/precomputed/metadata.py
nkemnitz: Probably not, no.

PR #333, opened 2020-04-26
Unpin Dependencies json5, backports-abc, and chardet

chrisroat:
backports-abc and chardet were pinned to their most recent versions, which are 3-4 years old (likely not to change much)
json5 was pinned to a 3 year old version, but development has continued with many minor releases in the interim.

Would the json5 update likely break anything that relied on older behavior?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
-backports-abc==0.5
+backports-abc>=0.5
 boto3>=1.4.7
-chardet==3.0.4
+chardet>=3.0.4
 compressed-segmentation>=1.0.0
 fastremap>=1.9.2
 fpzip>=1.1.3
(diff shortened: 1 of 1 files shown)


william-silversmith:
Hi Chris, just curious, are these dependency pins causing an issue for you? For what it's worth, json5 is almost never used. `json.loads` is called first, and json5 is only called if it fails (because json5 is extremely slow).

chrisroat:
I just noticed that my docker containers were downgrading json5 because of the dependency.  I investigated and saw these pinned deps read in by setup.py.   It's not a major problem, as my other pkgs aren't causing a conflict -- and my image size isn't a concern.

The travis failure looks to be a json error.  Not sure if it is related.

william-silversmith:
It looks like backports-abc isn't actually used anymore so we can just eliminate it. json5 looks like it's had some substantial work put into it so we should use the latest version.

PR #336, opened 2020-04-30
feat(ShardReader): add disassemble_shard to recover the input dict

william-silversmith:


william-silversmith:
Note: This refactor will break cached shard index on disk as it now stores it with the offset index.

PR #338, opened 2020-05-03
add_scale doesn't work for higher mips

manuel-castro:
Change add_scale fullres param from lowest mip to highest so add_scale will generate additional mips


william-silversmith:
I don't think this is broken. I think it's meant to be used like factor=(2,2,1) factor=(4,4,1) factor=(32,32,1) etc. This prevents issues like add_scale being by the same code over and over with the scale increasing infinitely.

manuel-castro:
Oh makes sense!

Issue #267, opened 2019-09-11
Read Sharded Meshes


william-silversmith:

william-silversmith:
![image (12)](https://user-images.githubusercontent.com/2517065/79049680-105cad80-7bf3-11ea-9fef-77eb5d40c25b.png)

Credit: Seun Ogedengbe

<img width="892" alt="image" src="https://user-images.githubusercontent.com/2517065/79049699-28ccc800-7bf3-11ea-9e40-0a7f74f47604.png">

Credit: Will Silversmith

william-silversmith:
This works for both graphene and precomputed now. Thanks @perlman!

PR #340, opened 2020-05-07
Support for neuroglancer multi-resolution sharded meshes

perlman:
This is an initial effort to read [neuroglancer](https://github.com/google/neuroglancer/)'s [multi-resolution sharded mesh format](https://github.com/google/neuroglancer/blob/master/src/neuroglancer/datasource/precomputed/meshes.md).

**TODO **

- [x] Request specific levels of detail
  - [x] ... and only request the relevant fragments over http
- [X]  [Convert](https://github.com/google/neuroglancer/blob/master/src/neuroglancer/datasource/precomputed/meshes.md#multi-resolution-mesh-fragment-data-file-format) from "stored model" space coordinates into correct model coordinates
- [X] Convert from "model space" into physical space
- [X] Concatenate meshes into a single mesh
- [X] Match function of other cloud-volume mesh functions
- [~] Check for missing manifests
  - (Via bulk prefetching of all manifests?)
  - [X] Added an exists method
- [ ] General code cleanup...

** Other issues (that probably won't be addressed) **
- read-only
  - Eventual write support would be done via unsharded storage & a compaction stage
- Only scale parameters in the homogenous transform are supported
-  Support for unsharded storage of multi-resolution meshes (much of the code would be similar, e.g. the mesh fragments are the same)
- Fuse meshes? (not sure how important this is)

Sample code
```python
import cloudvolume
test_ids = [ 710435991, 8392766771, 8393399736 ]
vol = cloudvolume.CloudVolume('https://storage.googleapis.com/fafb-ffn1-20190805/segmentation', mip=0, cache=False)
# Returns {'seg': Mesh}
for lod in reversed(range(4)):
   meshes = vol.mesh.get(test_ids, lod=lod, concat=True)
   for segment in meshes:
       mesh = meshes[segment]
       mesh.viewer()
```

### DIFF
--- a/cloudvolume/datasource/precomputed/mesh/__init__.py
+++ b/cloudvolume/datasource/precomputed/mesh/__init__.py
@@ -2,6 +2,7 @@ import os
 
 from .metadata import PrecomputedMeshMetadata
 from .unsharded import UnshardedLegacyPrecomputedMeshSource
+from .sharded import ShardedMultiLevelPrecomputedMeshSource
 
 
 from ..metadata import PrecomputedMetadata
@@ -16,8 +17,12 @@ class PrecomputedMeshSource(object):
     mesh_meta = PrecomputedMeshMetadata(meta, cache)
 
     if mesh_meta.is_sharded():
-      raise UnsupportedFormatError(red("This mesh volume is sharded. Mesh shard support is coming but is not available yet."))
-      # return ShardedLegacyPrecomputedMeshSource(mesh_meta, cache, config, readonly) 
+      if mesh_meta.info['sharding']['@type'] == 'neuroglancer_uint64_sharded_v1':
+        return ShardedMultiLevelPrecomputedMeshSource(mesh_meta, cache, config, readonly) 
+      else:
+        # @type = neuroglancer_legacy_mesh 
+        raise UnsupportedFormatError(red("This mesh volume is sharded. Legacy sharding is not available yet."))
+        # return ShardedLegacyPrecomputedMeshSource(mesh_meta, cache, config, readonly)
 
     return UnshardedLegacyPrecomputedMeshSource(mesh_meta, cache, config, readonly)
 
--- a/cloudvolume/datasource/precomputed/mesh/__init__.py
+++ b/cloudvolume/datasource/precomputed/mesh/__init__.py
@@ -2,7 +2,7 @@ import os
 
 from .metadata import PrecomputedMeshMetadata
 from .unsharded import UnshardedLegacyPrecomputedMeshSource
-from .sharded import ShardedMultiLevelPrecomputedMeshSource
+from .multilod import ShardedMultiLevelPrecomputedMeshSource, UnshardedMultiLevelPrecomputedMeshSource
 
 
 from ..metadata import PrecomputedMetadata
@@ -15,15 +15,15 @@ from ....cloudvolume import SharedConfiguration
 class PrecomputedMeshSource(object):
   def __new__(cls, meta, cache, config, readonly=False):
     mesh_meta = PrecomputedMeshMetadata(meta, cache)
-
-    if mesh_meta.is_sharded():
+    if mesh_meta.info.get('@type', None) == 'neuroglancer_multilod_draco':
       if mesh_meta.info['sharding']['@type'] == 'neuroglancer_uint64_sharded_v1':
+        # Sharded storage of multi-resolution mesh fragment data
         return ShardedMultiLevelPrecomputedMeshSource(mesh_meta, cache, config, readonly) 
       else:
-        # @type = neuroglancer_legacy_mesh 
-        raise UnsupportedFormatError(red("This mesh volume is sharded. Legacy sharding is not available yet."))
-        # return ShardedLegacyPrecomputedMeshSource(mesh_meta, cache, config, readonly)
-
+        # Unsharded storage of multi-resolution mesh fragment data
+        return UnshardedMultiLevelPrecomputedMeshSource()
+  
+    # Legacy single-resolution mesh format
     return UnshardedLegacyPrecomputedMeshSource(mesh_meta, cache, config, readonly)
 
   @classmethod
@@ -49,3 +49,5 @@ class PrecomputedMeshSource(object):
     meta = PrecomputedMetadata(cloudpath, cache, info={ 'mesh': mesh_dir })
 
     return PrecomputedMeshSource(meta, cache, config)
+
+
--- a/cloudvolume/datasource/precomputed/mesh/sharded.py
+++ b/cloudvolume/datasource/precomputed/mesh/sharded.py
@@ -1,9 +1,12 @@
+from collections import defaultdict
+
 import numpy as np
 
 from ..sharding import ShardingSpecification, ShardReader
 from ....storage import SimpleStorage
 from ....mesh import Mesh
-from ....lib import red
+from ....lib import yellow, red, toiter
+from .... import exceptions
 
 class ShardedMultiLevelPrecomputedMeshSource:
     def __init__(self, meta, cache, config, readonly=False):
@@ -20,7 +23,7 @@ class ShardedMultiLevelPrecomputedMeshSource:
         self.transform = np.array(self.meta.info['transform'] + [0,0,0,1]).reshape(4,4)
     
         if np.any(self.transform * np.array([[0,1,1,1],[1,0,1,1],[1,1,0,1],[1,1,1,0]])):
-            raise ValueError(red("Non-scale homogeneous transforms are not implemented"))
+            raise exceptions.MeshDecodeError(red("Non-scale homogeneous transforms are not implemented"))
 
     @property
     def path(self):
@@ -30,40 +33,43 @@ class ShardedMultiLevelPrecomputedMeshSource:
         """Fetch meshes at all levels of details.
 
         Parameters:
-        segids : int or [int ..]
-            Scalar or list of integer segment IDs.
+        segids: (iterable or int) segids to render
+
         lods : int, [int, ..] or None
             Level of detail(s) to retrieve.  0 is highest level of detail.
             None will fetch meshes at all levels.
 
         Returns:
-        [Mesh, ..] or [[Mesh, ..], ..]
-            A list of meshes for each ssegment ID is returned.
-            An additional layer is added when multiple lods are requested.
+        { segid: {lod: { Mesh, ... } } }
 
         Reference:
             https://github.com/google/neuroglancer/blob/master/src/neuroglancer/datasource/precomputed/meshes.md
         """
-        seg_list_return = True
-        if type(segids) in (int, float):
-            seg_list_return = False
-            segids = [ int(segids) ]
 
-        results = []
+        segids = toiter(segids)
+
+        # decode all the fragments
+        meshdata = defaultdict(list)
         for segid in segids:
             # Read the manifest (with a tweak to sharding.py to get the offset)
-            binary, shard_file_offset = self.reader.get_data(segid, self.meta.mesh_path, return_offset=True)
-            manifest = MultiLevelPrecomputedMeshManifest(binary)
+            result = self.reader.get_data(segid, self.meta.mesh_path, return_offset=True)
+            if result == None:
+                raise exceptions.MeshDecodeError(red(
+                    'Manifest not found for segment {}.'.format(segid)
+                ))
+            binary, shard_file_offset = result
+            manifest = MultiLevelPrecomputedMeshManifest(binary, segment_id=segid)
 
-            single_lod_return = False
-            if isinstance(lods, int):
-                single_lod_return = True
-                lods = [lods]
-            elif lods == None:
+            meshdata[segid] = defaultdict(list)
+
+            if lods == None:
                 lods = list(range(manifest.num_lods))
+            lods = toiter(lods)
 
             if any(lod >= manifest.num_lods for lod in lods):
-                raise ValueError(red("LOD value out of range (%d > %d)" % (max(lods), manifest.num_lods)))
+                raise exceptions.MeshDecodeError(red(
+                    'LOD value out of range ({} > {}) for segment {}.'.format(max(lods), manifest.num_lods, segid)
+                ))
 
             # Read the data for all LODs
             fragment_sizes = [ np.sum(lod_fragment_sizes) for lod_fragment_sizes in manifest.fragment_offsets ]
@@ -74,7 +80,6 @@ class ShardedMultiLevelPrecomputedMeshSource:
             full_path = self.reader.meta.join(self.reader.meta.cloudpath, self.path)
             stor =  SimpleStorage(full_path)
 
-            meshes = []
             for lod in lods:
                 lod_binary = stor.get_file(shard_file_name,
                         start=(shard_file_offset - total_fragment_size) + np.sum(fragment_sizes[0:lod]),
@@ -106,26 +111,17 @@ class ShardedMultiLevelPrecomputedMeshSource:
                     # Scale to native (nm) space
                     mesh.vertices =  mesh.vertices * (self.transform[0,0], self.transform[1,1], self.transform[2,2])
                     
-                    lod_meshes.append(mesh)
+                    meshdata[segid][lod].append(mesh)
 
-                # TODO: Fuse meshes before return?
-                meshes.append(lod_meshes)
-
-            results.append(meshes)
-
-        if single_lod_return:
-            results = results[0]
-        if seg_list_return:
-            return results
-        else:
-            return results[0]
+        return meshdata
 
 class MultiLevelPrecomputedMeshManifest:
     # Parse the multi-resolution mesh manifest file format:
     # https://github.com/google/neuroglancer/blob/master/src/neuroglancer/datasource/precomputed/meshes.md
     # https://github.com/google/neuroglancer/blob/master/src/neuroglancer/mesh/multiscale.ts
 
-    def __init__(self, binary):
+    def __init__(self, binary, segment_id):
+        self._segment = segment_id
         self._binary = binary
 
         # num_loads is the 7th word
@@ -159,7 +155,10 @@ class MultiLevelPrecomputedMeshManifest:
             offset += off_size
 
         # Make sure we read the entire manifest
-        assert(offset == len(binary))
+        if offset != len(binary):
+            raise exceptions.MeshDecodeError(red(
+                'Error decoding mesh manifest for segment {}'.format(segment_id)
+            ))
 
     @property
     def chunk_shape(self):
--- a/cloudvolume/datasource/precomputed/mesh/sharded.py
+++ b/cloudvolume/datasource/precomputed/mesh/sharded.py
@@ -1,0 +1,92 @@
+import numpy as np
+
+from ..sharding import ShardingSpecification, ShardReader
+from ....mesh import Mesh
+from ....lib import red
+
+class ShardedMultiLevelPrecomputedMeshSource(object):
+    def __init__(self, meta, cache, config, readonly=False):
+        self.meta = meta
+        self.cache = cache
+        self.config = config
+        self.readonly = bool(readonly)
+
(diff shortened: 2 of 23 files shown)


william-silversmith:
This is great Eric! Looking forward to seeing where this goes. :)

william-silversmith:
Linked to issue #267

perlman:
This code can now read the sharded meshes from both FAFB-FFN1 & FlyEM Hemibrain.

@william-silversmith: I added a quick hack to get the manifest location back with its contents, and am currently creating a new SimpleStorage for retrieving the meshes. How would you suggest reworking this to be a bit cleaner?

@william-silversmith: What should the mesh.get() function return for lod meshes? (The current return value is a list of fragments, one per lod.)

Much thanks to @jbms for fixing the clarification on a few key points in the [spec](https://github.com/google/neuroglancer/blob/master/src/neuroglancer/datasource/precomputed/meshes.md).

william-silversmith:
> [I] am currently creating a new SimpleStorage for retrieving the meshes. How would you suggest reworking this to be a bit cleaner?

I think this is fine for now. The multi-LOD spec departs from the ordinary sharded format so those departures would have to be stored somewhere... Perhaps a subclass of ShardReader would be slightly cleaner, but the performance cost of creating a new SimpleStorage is very low, the connection pool reuses connections. A ShardReader subclass would be able to avoid returning the offset from `get_data`, but it would also have to duplicate some code to maintain encapsulation. 

The major reworking I'd be looking forward to would be threaded retrieval. I've been working on making that possible for ShardReader.exists, and the techniques can be ported to ShardReader.get_data. Once that happens, we can dramatically increase the performance here.

>  What should the mesh.get() function return for lod meshes? (The current return value is a list of fragments, one per lod.)

My feeling is that retrieving all the LODs all the time is kind of a waste of the benefits of the format. What about adding a `get(segids, lod=scalar or list)` flag? Similarly, for fusing, you can fuse the meshes retrieved within a given LOD. If you wanted to get a bit more sophisticated, perhaps we could have a way of retrieving meshes with respect to a point such that it cobbles together a cohesive view of the final mesh based on distance to that point similarly to neuroglancer. 

Thanks for your excellent work! I like your code too Eric. <3 :) I'll give it a try a bit later today.

perlman:
> A ShardReader subclass would be able to avoid returning the offset from get_data, but it would also have to duplicate some code to maintain encapsulation.

Makes sense. Maybe we can put half of that work into a private method in SharedReader, which can then be called from the subclass? (The index lookup is an example of code I wouldn't want to be replicated.)

> My feeling is that retrieving all the LODs all the time is kind of a waste of the benefits of the format. What about adding a get(segids, lod=scalar or list) flag?

Done (and each LOD is downloaded as it's own range request). Default is still all meshes defined in the mesh manifest.

> Similarly, for fusing, you can fuse the meshes retrieved within a given LOD.

Next step. :)

> If you wanted to get a bit more sophisticated, perhaps we could have a way of retrieving meshes with respect to a point such that it cobbles together a cohesive view of the final mesh based on distance to that point similarly to neuroglancer.

I'll leave this to someone else.  ;-) I'd be content with a function to do mesh selection by resolution (e.g., I want the best mesh for 32 x 32 z 40).

cloudvolume/datasource/precomputed/mesh/sharded.py
```diff
@@ -0,0 +1,199 @@
...
+        Returns:
+        [Mesh, ..] or [[Mesh, ..], ..]
+            A list of meshes for each ssegment ID is returned.
+            An additional layer is added when multiple lods are requested.
+
+        Reference:
+            https://github.com/google/neuroglancer/blob/master/src/neuroglancer/datasource/precomputed/meshes.md
+        """
+        seg_list_return = True
+        if type(segids) in (int, float):
```
william-silversmith, line 48: You can use `cloudvolume.lib.toiter` here: `segids = toiter(segids)`

cloudvolume/datasource/precomputed/mesh/sharded.py
```diff
@@ -0,0 +1,199 @@
...
+
+            # Read fragment sizes
+            off_size = 4 * self.num_fragments_per_lod[lod]
+            self._fragment_offsets.append(
+                np.frombuffer(self._binary[offset:offset + off_size], dtype=np.uint32)
+            )
+            offset += off_size
+
+        # Make sure we read the entire manifest
+        assert(offset == len(binary))
```
william-silversmith, line 162: I usually prefer not to have asserts in production. Maybe create a new exception or reuse  `cloudvolume.exceptions.MeshDecodeError`?

cloudvolume/datasource/precomputed/mesh/sharded.py
```diff
@@ -0,0 +1,199 @@
...
+            total_fragment_size = np.sum(fragment_sizes)
+
+            # Kludge to hijack sharding.py to read the data
+            shard_file_name = self.reader.get_filename(segid)
+            full_path = self.reader.meta.join(self.reader.meta.cloudpath, self.path)
+            stor =  SimpleStorage(full_path)
+
+            meshes = []
+            for lod in lods:
+                lod_binary = stor.get_file(shard_file_name,
```
william-silversmith, line 79: Ideally, you could use `Storage.get_files` to extract all the LOD binaries at once so you don't take the performance hit. Otherwise, it's conceivably faster to do what you were doing before and just discard the extra LODs. I think it's also fine to not worry about performance right now and just get something working out. I'll have to tune this for performance later anyway.

cloudvolume/datasource/precomputed/mesh/sharded.py
```diff
@@ -0,0 +1,199 @@
...
+            segids = [ int(segids) ]
+
+        results = []
+        for segid in segids:
+            # Read the manifest (with a tweak to sharding.py to get the offset)
+            binary, shard_file_offset = self.reader.get_data(segid, self.meta.mesh_path, return_offset=True)
+            manifest = MultiLevelPrecomputedMeshManifest(binary)
+
+            single_lod_return = False
+            if isinstance(lods, int):
```
william-silversmith, line 59: For this, you can also use `toiter(lods, is_iter=True)` to get both an iterable lods and whether the input was a scalar.

cloudvolume/datasource/precomputed/mesh/sharded.py
```diff
@@ -0,0 +1,199 @@
...
+
+        Parameters:
+        segids : int or [int ..]
+            Scalar or list of integer segment IDs.
+        lods : int, [int, ..] or None
+            Level of detail(s) to retrieve.  0 is highest level of detail.
+            None will fetch meshes at all levels.
+
+        Returns:
+        [Mesh, ..] or [[Mesh, ..], ..]
```
william-silversmith, line 40: In other Mesh sources, I believe we return `{ segid: Mesh or [ Meshes ], ... }`

william-silversmith:
> Maybe we can put half of that work into a private method in SharedReader, which can then be called from the subclass?

Sure! If you're up for it, give it a shot. If it gets awkward though, don't worry too much about the duplication of logic. I like to follow the "rule of three" where I'll sometimes let some code be duplicated until the third use so the commonalities become more clear. Just do whatever looks most elegant to you, which might include keeping the offset parameter.

> >  If you wanted to get a bit more sophisticated, perhaps we could have a way of retrieving meshes with respect to a point such that it cobbles together a cohesive view of the final mesh based on distance to that point similarly to neuroglancer.
> 
> I'll leave this to someone else. ;-) I'd be content with a function to do mesh selection by resolution (e.g., I want the best mesh for 32 x 32 z 40).

That's fine, I was just spitballing. It's possible that feature is really only useful for visualization. 

Thanks again for all the work you put into this Eric!

cloudvolume/datasource/precomputed/mesh/sharded.py
perlman: > My feeling is that retrieving all the LODs all the time is kind of a waste of the benefits of the format.

... is what is being implemented here. Maybe a minor optimization would be to request adjacent LODs in a single request. Still, this change is good change for any request that does not include the full resolution mesh. (Note that all of the fragments for a given lod are retrieved in this call.)

cloudvolume/datasource/precomputed/mesh/sharded.py
perlman: I may have overloaded this function here: scalar segid & scalar lod -> [ Mesh fragments], one being scalar one being a list -> [[Mesh fragments]], both being lists -> [[[Mesh fragments]]].  Eeek!

perlman: Current return is now a simple  { segid: {lod: { Mesh, ... } } }.... I don't love it though.

william-silversmith:
Good news bad news time! Good news: I parallelized get_data so now it should be possible to dramatically speed this function up. Bad news: I suppose we need to re-integrate the return_offset flag.

william-silversmith:
Just to be clear, don't worry about it. Just let me know when you're done with the PR and I can fix it before merging.

cloudvolume/datasource/precomputed/mesh/sharded.py
perlman: ... and now it's { lod: { segid: Mesh } }.

perlman:
> Good news bad news time! 
🤣 

I think I've tackled all of the small things _except_ the interfaces for reading data.

As-is, this code should now more accurately reflect the other mesh functions, and will serve the purpose most people want: downloading complete meshes of a given level of detail. There's clearly room to improve, e.g. bounding boxes or dynamic based on a view point, neither of which are particularly urgent.

> Just to be clear, don't worry about it. Just let me know when you're done with the PR and I can fix it before merging.

I think I'm there. My tweaks are increasingly minor.

perlman:
@william-silversmith: This is updated to work with your latest code.

https://github.com/seung-lab/cloud-volume/blob/master/cloudvolume/datasource/precomputed/sharding.py#L511 was erroring ("TypeError: 'list' object is not an iterator").

I would still love to find a clean way to avoid both doing the double-lookup of the manifest info _and_ avoid making a complete copy of get_data.  Thoughts?

The relevant lines are https://github.com/perlman/cloud-volume/blob/sharded_mesh/cloudvolume/datasource/precomputed/mesh/sharded.py#L49-L50

cloudvolume/datasource/precomputed/mesh/__init__.py
```diff
@@ -16,8 +17,12 @@ def __new__(cls, meta, cache, config, readonly=False):
     mesh_meta = PrecomputedMeshMetadata(meta, cache)
 
     if mesh_meta.is_sharded():
-      print(red("This mesh volume is sharded. Mesh shard support is coming but is not available yet."))
-      # return ShardedLegacyPrecomputedMeshSource(mesh_meta, cache, config, readonly) 
+      if mesh_meta.info['sharding']['@type'] == 'neuroglancer_uint64_sharded_v1':
+        return ShardedMultiLevelPrecomputedMeshSource(mesh_meta, cache, config, readonly) 
+      else:
+        # @type = neuroglancer_legacy_mesh 
+        raise UnsupportedFormatError(red("This mesh volume is sharded. Legacy sharding is not available yet."))
```
william-silversmith, line 24: I'm not sure how to handle this case because raising an error means the images and skeletons are also inacessible.

perlman: I had not thought about that.  Maybe we can create a legacy sharded mesh class that does nothing but throw exceptions if it's used?

william-silversmith:
That sounds like a much better solution!

cloudvolume/datasource/precomputed/mesh/__init__.py
william-silversmith: This is a much better idea!

william-silversmith:
This is looking pretty good! We can always brush up the performance aspect later. Would you mind writing some test(s) for it? If you point me to a particular shard file, I can upload it to gs://seunglab-test/ so it can be forever tested. 

It doesn't need to be anything too crazy, an integration test is sufficient.

cloudvolume/datasource/precomputed/mesh/__init__.py
perlman: I created a placeholder UnshardedMultiLevelPrecomputedMeshSource in the now-renamed multilod.py. It does little but error when you try to do anything.

perlman:
> This is looking pretty good! We can always brush up the performance aspect later. Would you mind writing some test(s) for it? If you point me to a particular shard file, I can upload it to gs://seunglab-test/ so it can be forever tested.
> 
> It doesn't need to be anything too crazy, an integration test is sufficient.

This sounds like a good idea. I don't have have the ability to generate any data -- would it make sense just to pull one of the shard files from any of the public neuroglancer datasets? (e.g. hemibrain).

william-silversmith:
Yea that's perfect! Just use a public dataset.

perlman:
This now includes a minimal test case, with a reference to the files accessed. When I have chance I can try to pick some good & bad IDs that all hash to the same shard file.

william-silversmith:
This is hella cool. Thank you so much for your work!!

jefferis:
Hurrah! Thank you both very much (especially @perlman in this case!).

william-silversmith:
I'm going to make a new release sometime today. :)

perlman:
I am sure there will be follow up PRs to fix bugs & add other useful interfaces (like a bbox() defined by the oct-tree regions).  I'll see what pops up with @jefferis and others use it. ;)

jefferis:
@perlman @william-silversmith: a couple of queries for one or both of you after using this in earnest a bit.
1. In contrast (I think) to some other mesh formats, the get method cannot accept strings (only ints). Would it be hard to change this? I often find 64 bit int support fragile across programming languages (e.g. base R only has int32, so r to python calls doesn't support int64). 
2. Again some of the other mesh sources I have used (flywire) have a save method. Here there is none possibly because of the issue of multi-lod meshes. Is it worth adding? Alternatively is there a general pattern for saving (e.g. in obj) that is guranteed to work across mesh formats?

fcollman:
I'll add my two cents on point 2.  I think once you have vertices and faces there are a lot of mesh libraries out there for saving meshes.  I think of cloud-volume as the interface for getting/saving highly distributed files for cloud based connectomics.  If you want to save versions of meshes as simple files, I think there are existing packages that cover a large range of file formats.  Meshparty for example uses trimesh, which enables it to save in obj (or ply, or many others).

So in that case meshparty calls cloud-volume to get a mesh, uses that to instantiate a class that is derived from a trimesh base class.  We can then use that to save meshes in a variety of formats, though as we develop more analytic layers on top of the mesh we have created an hdf based format to store these extra analytics.  One immediate relevant example, is a place to store the extra edges that link disconnected portions of proofread meshes.  Others would include skeleton representations and the mappings between the skeleton and the mesh.

jefferis:
> I'll add my two cents on point 2. I think once you have vertices and faces there are a lot of mesh libraries out there for saving meshes.
Thanks Forrest. That seems absolutely reasonable.

It's just that I wrote some wrapper code to fetch/save meshes from FlyWire and I was imagining that it would just work when I switched to hemibrain as a source, but it doesn't because the interface that cloudvolume presents seems to vary according to the type of mesh source – and to me that was a bit surprising. So it would be good to know if there is a recommended pattern for fetch+save.

fcollman:
agreed about the install process, we have work to do there.  Mostly I think by making more dependancies optional and writing better cross-platform instructions. 

regarding interfaces, I can see how the LOD dictionary is both necessary and confusing. I would agree with you that a more uniform looking interface is a better design ideal.  For example, making the highest LOD the default, and only returning that by default, and allowing people to specify LOD, or specify that they actually want the full dictionary.

william-silversmith:
@jefferis It looks like this was an oversight. I'll make sure to add save back in there. The implementation of save is very simple and worth understanding: 
```
def save(self, segids, filepath=None, file_format='ply'):
    segids = toiter(segids)

    mesh = self.get(segids, fuse=True, remove_duplicate_vertices=True)

    if file_format == 'obj':
      data = mesh.to_obj()
    elif file_format == 'ply':
      data = mesh.to_ply()
    elif file_format == 'precomputed':
      data = mesh.to_precomputed()
    else:
      raise NotImplementedError('Only .obj, .ply, and precomputed are currently supported.')

    if not filepath:
      filepath = str(segids[0]) + "." + file_format
      if len(segids) > 1:
        filepath = "{}_{}.{}".format(segids[0], segids[-1], file_format)

    try:
      filepath.write(data)
    except AttributeError:
      with open(filepath, 'wb') as f:
        f.write(data)
```

As you can see, each Mesh object comes equipped with a way to convert it into an obj or ply byte string. You can then do what you will with it. I've often thought that the `save` function is somewhat superfluous and too tightly coupled with `get`, but it is convenient for someone without a deep understanding of CloudVolume.

perlman:
> In contrast (I think) to some other mesh formats, the get method cannot accept strings (only ints). Would it be hard to change this? I often find 64 bit int support fragile across programming languages (e.g. base R only has int32, so r to python calls doesn't support int64).

I'll let @william-silversmith chime in. uint64 are used in many functions, all of which would need to add the hook to convert. Would it make sense to keep a small wrapper, in python, with your R code for doing the string to uint64 conversion?

> It's just that I wrote some wrapper code to fetch/save meshes from FlyWire and I was imagining that it would just work when I switched to hemibrain as a source, but it doesn't because the interface that cloudvolume presents seems to vary according to the type of mesh source – and to me that was a bit surprising. So it would be good to know if there is a recommended pattern for fetch+save.

https://github.com/seung-lab/cloud-volume/blob/master/cloudvolume/datasource/precomputed/mesh/multilod.py#L72
vs 
https://github.com/seung-lab/cloud-volume/blob/master/cloudvolume/datasource/graphene/mesh/unsharded.py#L137

With the exception of fusing (not implemented yet), the return "should" be the same`{ segid: Mesh }.`

Please send me a snippit of how you are accessing the meshes and I'll see what I can do. (It's quite likely I broke something in my continual tweaking.)

> Alternatively is there a general pattern for saving (e.g. in obj) that is guranteed to work across mesh formats?

I am shoving the data into a standard cloud-volume Mesh object, which ought to behave the same as the others...  Maybe there is an issue with the lack of normals?

> regarding interfaces, I can see how the LOD dictionary is both necessary and confusing. I would agree with you that a more uniform looking interface is a better design ideal. For example, making the highest LOD the default, and only returning that by default, and allowing people to specify LOD, or specify that they actually want the full dictionary.

@fcollman: This is exactly what I converged on with the 'final' version. If there is a use case for getting _all_ LODs I'd implement a separate function use a more efficient range request to fetch them all at once.

perlman:
> @jefferis It looks like this was an oversight. I'll make sure to add save back in there. The implementation of save is very simple and worth understanding

Aha! This function was never on my radar as something to implement.

Should it come for free in a base class?

It looks like failed at having UnshardedLegacyPrecomputedMeshSource extend any existing class...

william-silversmith:
I've been using `UnshardedLegacyPrecomputedMeshSource` as a base class in a few places now, but the more places we use it the more worried I become about the inheritance structure becoming brittle and hard to reason about. Adding it as a base class will fix this case though.

perlman:
> I've been using UnshardedLegacyPrecomputedMeshSource as a base class in a few places now, but the more places we use it the more worried I become about the inheritance structure becoming brittle and hard to reason about. Adding it as a base class will fix this case though.

It sounds like a BaseMeshSource would be useful? I'm not entirely sure which functions should be defined there, but it would go a long way towards having a standard interface for mesh access.

william-silversmith:
That sounds about right. At the very least the save function should be
defined in the base class.

jefferis:
thanks everyone for additional info and @william-silversmith thanks for the save example.

Just a note that for my own libraries, for user facing code equivalent to this:

```
def save(self, segids, filepath=None, file_format='ply'):
    segids = toiter(segids)
```

I have a function that might replace `toiter`, which converts reasonable inputs (including string and numeric) to int64 ids with some basic validation checks. But my libraries are mostly intended for end users whereas I guess CloudVolume is a mix of developer/end user targeting.

Issue #349, opened 2020-06-11
Hosting precomputed volume on Windows OS


austinhoag:
Hi Will,

A member of our lab is trying to host some local precomputed volumes on their Windows machine so they can view them in Neuroglancer. We are stuck trying to figure out the correct syntax on Windows. I have limited Windows experience, so it could boil down to that.  

The layer directory (that contains the info file) we want to host is:
`C:\Users\jverpeut\Documents\neuroglancer\rawdata_an21`
So we have tried:
```python
from cloudvolume import CloudVolume
vol=CloudVolume(r'file://C:\Users\jverpeut\Documents\neuroglancer\rawdata_an21')

cloudvolume.exceptions.InfoUnavailableError: No info file was found: file://C:\Users\\Users\jverpeut\Documents\neuroglancer\rawdata_an21\info
```

We verified that the path to the layer dir is correct in python and contains an info file:
```python
os.listdir(r'C:\Users\jverpeut\Documents\neuroglancer\rawdata_an21')
['10000_10000_10000', '20000_20000_10000', '40000_40000_10000', '5000_5000_10000', '80000_80000_10000', 'info', 'provenance']
```

I have copied this layer over to linux and I can host it fine on there. 

What are we missing here?

william-silversmith:
Hi Austin,

The problem is probably related to something in the path handling of CloudVolume or Storage. I don't have regular access to a Windows machine so I depend on others to submit patches when they have problems though when I have access to a machine I try to help. Unfortunately, due to COVID and no access to the lab, I don't even have the Windows 10 VM I used to tool around with. 

If you get a chance, try seeing what the inputs are when it tries to fetch the `info` file in https://github.com/seung-lab/cloud-volume/blob/master/cloudvolume/datasource/precomputed/metadata.py#L194.

Sorry about that...

Will

PR #350, opened 2020-06-16
address issue with windows basepath extraction

austinhoag:
To address: [https://github.com/seung-lab/cloud-volume/issues/349](https://github.com/seung-lab/cloud-volume/issues/349), an issue where windows paths were not parsed properly. Specifically, the bucket was getting added to the cloudpath twice.   

I added a simple test to make sure the fix works.

### DIFF
--- a/cloudvolume/paths.py
+++ b/cloudvolume/paths.py
@@ -117,7 +117,6 @@ def extract(cloudpath, windows=None, disable_toabs=False):
   if len(cloudpath) == 0:
     return ExtractedPath('','','','','','','')
 
-  # windows_file_re = re.compile(r'((?:\w:\\)[\d\w_\.\-]+(?:\\)?)') # for C:\what\a\great\path
   windows_file_re = re.compile(r'((?:\w:\\)[\d\w_\.\-]+)')
   bucket_re = re.compile(r'^(/?[~\d\w_\.\-]+(?::\d+)?)/') # posix /what/a/great/path
   
--- a/cloudvolume/paths.py
+++ b/cloudvolume/paths.py
@@ -161,8 +161,12 @@ def extract(cloudpath, windows=None, disable_toabs=False):
     no_bucket_basepath = split_char.join(splitties[1:-1])
   elif len(splitties) >= 2:
     dataset, layer = splitties[-2:]
-    no_bucket_basepath = split_char.join(splitties[1:-1])
-    basepath = split_char.join([bucket] + splitties[1:-1])
+    if windows:
+      no_bucket_basepath = split_char.join(splitties[2:-1])
+      basepath = split_char.join([bucket] + splitties[2:-1])
+    else:
+      no_bucket_basepath = split_char.join(splitties[1:-1])
+      basepath = split_char.join([bucket] + splitties[1:-1])
 
   return ExtractedPath(
     fmt, protocol, bucket, 
--- a/cloudvolume/paths.py
+++ b/cloudvolume/paths.py
@@ -132,8 +132,7 @@ def extract(cloudpath, windows=None, disable_toabs=False):
     abspath = toabs    
 
   fmt, protocol, cloudpath = extract_format_protocol(cloudpath)
-  print("cloudpath:")
-  print(cloudpath)
+
   split_char = '/'
   if protocol == 'file':
     cloudpath = abspath(cloudpath)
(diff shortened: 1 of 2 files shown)


test/test_paths.py
```diff
@@ -140,3 +140,13 @@ def okgoogle(url):
...
   assert path.layer == 'layer'
+
+
+def test_windows_path_extraction():
+  """ This test is added specifically to make sure
+  that no_bucket_basepath does not contain the bucket,
+  for windows paths. """
+  extract = paths.extract(r'file://C:\wow\this\is\a\cool\path', windows=True, disable_toabs=True)
+  assert extract.protocol == 'file'
+  assert extract.bucket == 'C:\\wow\\'
```
william-silversmith, line 151: Hmm, I'm not sure automated testing will give the correct result because `split_char` is set by `os.path.sep`.

william-silversmith: Maybe it's worth redefining this part of the code:

```  split_char = '/'
  if protocol == 'file':
    cloudpath = abspath(cloudpath)
    if windows:
      bucket_re = windows_file_re
      split_char = '\\'```

Comment (austinhoag) [reply test/test_paths.py:151]: Hmm, I'm not sure I follow your point. 

That part of the test is actually from the other test that already existed in this file. I only added the line:
```python
assert 'wow' not in extract.no_bucket_basepath
```
everything else was copy paste of the top of the previous test.

In any case the tests pass on linux and windows platforms, so I'm not sure why it wouldn't pass automated testing. The split_char is set to `'\\'` in the test because `windows=True` in:
```python
extract = paths.extract(r'file://C:\wow\this\is\a\cool\path', windows=True, disable_toabs=True)
```
Maybe there is something I don't understand about the automated testing. Can you explain what you mean?

william-silversmith:
In the extract paths code, there's a call to os.path.sep I don't fully
understand why the windows tests pass on linux when os.path.sep will return
/ instead of \

william-silversmith:
To be clear, this is a preexisting issue, not one you added.

austinhoag:
Oh, I see what you mean. I mistook the code block you had written as the actual code, not a suggestion. You are right that the separator was set by `os.path.sep` which on linux is `/`. It was lucky that the test passed on linux. It was because `splittlies` was a list of length 1, where the single entry was the entire cloudpath, such that the following code block was getting executed:
```python
 elif len(splitties) == 1:
    dataset = bucket
    layer = splitties[0]
    basepath = split_char.join(splitties[:-1])
    no_bucket_basepath = split_char.join(splitties[1:-1])
```
so `no_bucket_basepath` was getting set to an empty string, so my assert statement:
`assert 'wow' not in no_bucket_basepath` was true by accident.

I added your suggested change, and I also had to change the windows regex slightly to get the bucket and basepath to come out correctly for windows paths. Tests should now pass for the right reason on linux.

cloudvolume/paths.py
```diff
@@ -117,7 +117,8 @@ def extract(cloudpath, windows=None, disable_toabs=False):
   if len(cloudpath) == 0:
     return ExtractedPath('','','','','','','')
 
-  windows_file_re = re.compile(r'((?:\w:\\)[\d\w_\.\-]+(?:\\)?)') # for C:\what\a\great\path
+  # windows_file_re = re.compile(r'((?:\w:\\)[\d\w_\.\-]+(?:\\)?)') # for C:\what\a\great\path
+  windows_file_re = re.compile(r'((?:\w:\\)[\d\w_\.\-]+)')
```
william-silversmith, line 121: `(?:\\)?)` allows the paths to end with a slash

austinhoag: If `(?:\\)?)` is included, `bucket` gets set to `C:\\wow\\` when it should be `C:\\wow` right? So when basepath gets created like:
```python
basepath = split_char.join([bucket] + splitties[1:-1])
```
It has too many `\\` characters between bucket and the rest of the path:
`C:\\wow\\\\this\\is\\a\\cool`

With the updated regex, a cloudpath ending with a `\\` will still be interpreted correctly in windows. I added a check for this in the new test function.

william-silversmith: I see. Would you mind removing the commented out line? Thanks for your help!

austinhoag: sure.

PR #355, opened 2020-07-03
adding smart graphene multi-level mesh stitching

fcollman:
I have implemented a more elaborate mesh stitching routine for graphene sharded.  It now takes into account the mesh level of each fragment that is downloaded, in order to label the chunk boundaries dynamically on each fragment in a way that scales with the level of the chunk the fragment came from, and then only removes duplicate vertices on those chunk boundaries.

To do so, i did a little refactoring of the deduplication code so i could call it outside of the Mesh class.  Happy to consider refactoring this in some way, and might be duplicating effort that exists elsewhere in pychunkedgraph. 

I've tested that this enables proper mesh stitching of graphene sharded meshes on a proofread ID example from minnie, but haven't tested it on fly.

### DIFF
--- a/cloudvolume/frontends/graphene.py
+++ b/cloudvolume/frontends/graphene.py
@@ -21,6 +21,7 @@ from ..volumecutout import VolumeCutout
 from ..datasource.graphene.metadata import GrapheneApiVersion
 
 from .precomputed import CloudVolumePrecomputed
+import tqdm
 
 def warn(text):
   print(colorize('yellow', text))
@@ -36,7 +37,7 @@ def to_unix_time(timestamp):
   if isinstance(timestamp, datetime): # NB. do not change to elif
     timestamp = datetime.timestamp(timestamp)
 
-  if not isinstance(timestamp, int) and timestamp is not None:
+  if not isinstance(timestamp, float) and timestamp is not None:
     raise ValueError("Not able to convert {} to UNIX time.".format(timestamp))
 
   return timestamp
--- a/cloudvolume/frontends/graphene.py
+++ b/cloudvolume/frontends/graphene.py
@@ -37,10 +37,10 @@ def to_unix_time(timestamp):
   if isinstance(timestamp, datetime): # NB. do not change to elif
     timestamp = datetime.timestamp(timestamp)
 
-  if not isinstance(timestamp, float) and timestamp is not None:
+  if not isinstance(timestamp, (int, float, np.integer, np.floating)) and timestamp is not None:
     raise ValueError("Not able to convert {} to UNIX time.".format(timestamp))
 
-  return timestamp
+  return int(timestamp)
 
 class CloudVolumeGraphene(CloudVolumePrecomputed):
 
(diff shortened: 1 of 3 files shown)


fcollman:
i think one of the travis builds spontaneously crashed

cloudvolume/frontends/graphene.py
```diff
@@ -36,7 +37,7 @@ def to_unix_time(timestamp):
   if isinstance(timestamp, datetime): # NB. do not change to elif
     timestamp = datetime.timestamp(timestamp)
 
-  if not isinstance(timestamp, int) and timestamp is not None:
+  if not isinstance(timestamp, float) and timestamp is not None:
```
william-silversmith, line 40: This should read `isinstance(timestamp, (int, float))` and should return `int(timestamp)` I think. Using a UNIX timestamp directly is supported and will usually be regarded as an int.

william-silversmith:
The `levels` information is easily extracted from the segment ID on a mesh. Wouldn't it be cleaner in the function signature to ensure that the segid is attached properly to the meshes and then extract the levels dynamically as needed?

fcollman:
it would, but the the manifest for the sharded meshes doesn't return the seg_id, it only returns the level/shardfile, so I can't attach the seg_id to the meshes calculated like that.

william-silversmith:
Let me see if @akhileshh would be willing to add in the segid to the manifest.

william-silversmith:
I forgot that I already asked him to add an option for adding in segids for the bypass mode. Let me investigate that...

william-silversmith:
The "return_segids" option looks like the example below. There's a lot of redundant information. The problem is that the segids are present, but its possible for multiple segids to map to a single shard but no way to disambiguate the byte range. I'm asking @akhileshh if the order of fragments and seigds exactly matches.

```
{
  "fragments": [
    "~3/47461891-0.shard:3430219:1073", 
    "~3/45385220-0.shard:310763:1182", 
    "~3/42020357-0.shard:2270331:304", 
    "~3/46722564-0.shard:1920478:580", 
   ...
   ],
   "seg_ids": [
       241653690820204416, 
    240538786566506912, 
    238732289498941469, 
    241256767659446532, 
    240846100603342100, 
    241545115120699551, 
    241544290486978899, 
    241544290486978944, 
]
}
```

```

akhileshh:
The same segids are used to generate the fragments so they share the same order.
Neuroglancer doesn't need segids returned at the moment so it's optional as a query parameter.

william-silversmith:
Thanks Akhilesh! @fcollman would you want to make these changes? I can make them for you if it's bothersome.

fcollman:
I'm running into a very heavy meeting week, so i'm not sure whether i'm going to have time in the next few days to get this done, so if you beat me to it just push to this branch and i'll see it.

william-silversmith:
Refactored to assign a segid to every mesh and refactored to use CloudFiles.

william-silversmith:
I tested this on minnie3_v1 with segid 864691136882099822 which doesn't appear to have self touches. I still saw duplicate vertices. :-/

fcollman:
k, i'll take a look, but whats' most important is that the mesh is connected properly, there are ways that duplicate vertices can occur hyper locally due to draco compression that are not problematic in terms of mesh topology properly stitching.  As long as the duplicate vertices are merged at seam boundaries (and no more than that) the code is working properly.

PR #362, opened 2020-07-14
fix: special logic for handling slices

### LINKED ISSUE #314: Writing to volume (by slicing notation) is not properly checking slice.start < slice.stop
cv[(slice(0,1),slice(0,1),slice(779,344))] = 0  appears to clear out 344:779 rather than the expected behavior of not writing anything because the effective range is empty.

william-silversmith:
Resolves #314

### DIFF
--- a/cloudvolume/lib.py
+++ b/cloudvolume/lib.py
@@ -403,9 +403,13 @@ class Bbox(object):
         slices, bounded=bounded, autocrop=autocrop
       )
 
+    for slc in slices:
+      if slc.step not in (None, 1):
+        raise ValueError("Non-unitary steps are unsupported. Got: " + str(slc.step))
+
     return Bbox(
       [ slc.start for slc in slices ],
-      [ (slc.start if slc.stop < slc.start and (slc.step is None or slc.step > 0) else slc.stop) for slc in slices ]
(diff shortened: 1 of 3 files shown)


william-silversmith:
@shangmu what do you think of this fix? Originally I wanted people to specify bbox via any corners of the bounding box, but it seems like slice notation has additional semantics that make this approach non-sensical, so I added special casing for it.

cloudvolume/lib.py
```diff
@@ -405,7 +405,7 @@ def from_slices(cls, slices, context=None, bounded=False, autocrop=False):
 
     return Bbox(
       [ slc.start for slc in slices ],
-      [ slc.stop for slc in slices ]
+      [ (slc.start if slc.stop < slc.start and (slc.step is None or slc.step > 0) else slc.stop) for slc in slices ]
```
shangmu, line 408: I would rather throw an error saying this is unsupported if a step is ever specified. With general convention, people would have expected an subset of the sections/pixels in the case of |step|>1 and sections/pixels in the reverse order in the case of step<0

william-silversmith: Okay, I'll raise a ValueError.

PR #363, opened 2020-07-16
fix(encode_jpeg): Neuroglancer expects vertical tile layout

nkemnitz:
Don't exactly understand the operations involved here, but I verified that with the change below a chunk downloaded with CloudVolume and re-uploaded as JPEG shows up at the correct position / in the correct orientation in Neuroglancer. Previously, for my 128x128x16 chunks Neuroglancer would complain about shape 16384x16, instead of the expected 128x2048

### DIFF
--- a/cloudvolume/chunks.py
+++ b/cloudvolume/chunks.py
@@ -90,7 +90,7 @@ def encode_jpeg(arr):
 
   reshaped = arr.T 
   reshaped = np.moveaxis(reshaped, 0, -1)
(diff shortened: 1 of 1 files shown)


william-silversmith:
Rebased as I added some automated tests for JPEG encoding.

william-silversmith:
Just curious, how did you stumble upon this? Just trying stuff?

nkemnitz:
Tried to view my first jpeg encoded image shard and Neuroglancer complained about the chunk shape.

Issue #354, opened 2020-06-23
Error fetching sharded data


schlegelp:
Hi,

I'm trying to load some of the Janelia hemibrain EM image data but run into this issue:

```Python
>>> vol = CloudVolume('precomputed://gs://neuroglancer-janelia-flyem-hemibrain/emdata/clahe_yz/jpeg', fill_missing=True)
>>> vol[6300:6500, 20400:20600:, 14000]
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-59-f9e19f62db12> in <module>
----> 1 vol[6300:6500, 20400:20600:, 14000]

~/.local/lib/python3.7/site-packages/cloudvolume/frontends/precomputed.py in __getitem__(self, slices)

~/.local/lib/python3.7/site-packages/cloudvolume/frontends/precomputed.py in download(self, bbox, mip, parallel, segids, preserve_zeros, agglomerate, timestamp, stop_layer)

~/.local/lib/python3.7/site-packages/cloudvolume/datasource/precomputed/image/__init__.py in download(self, bbox, mip, parallel, location, retain, use_shared_memory, use_file, order)

~/.local/lib/python3.7/site-packages/cloudvolume/datasource/precomputed/sharding.py in from_dict(cls, vals)

TypeError: __init__() missing 1 required positional argument: 'data_encoding' 
```

Any idea what I'm doing wrong here?

Thanks! :)

schlegelp:
Forgot to add: this is with cloudvolume `1.18.1`

schlegelp:
With more traceback:
```Python
<ipython-input-2-dbdeda5c0ce0> in <module>
----> 1 vol[6300:6500, 20400:20600:, 14000]

~/.pyenv/versions/3.7.5/lib/python3.7/site-packages/cloudvolume/frontends/precomputed.py in __getitem__(self, slices)
    526     requested_bbox = Bbox.from_slices(slices)
    527 
--> 528     img = self.download(requested_bbox, self.mip)
    529     return img[::steps.x, ::steps.y, ::steps.z, channel_slice]
    530 

~/.pyenv/versions/3.7.5/lib/python3.7/site-packages/cloudvolume/frontends/precomputed.py in download(self, bbox, mip, parallel, segids, preserve_zeros, agglomerate, timestamp, stop_layer)
    568       parallel = self.parallel
    569 
--> 570     img = self.image.download(bbox, mip, parallel=parallel)
    571 
[...]

Issue #221, opened 2019-06-03
Support of new Sharded Neuroglancer Format for Image Data


unidesigner:
I was trying to pull a cut-out from the new Google segmentation data of FAFB using the following code

```
from cloudvolume import CloudVolume
vol = CloudVolume("https://storage.googleapis.com/fafb-ffn1-20190521/segmentation")
```

but I get a voxel_offset error. I think this is related to the new [Sharded Format](https://github.com/google/neuroglancer/blob/master/src/neuroglancer/datasource/precomputed/README.md#sharded-format) not yet being supported by cloud-volume.
Is there a plan to support this new format in the library?

@jbms: Is the new format going to be supported by the new TensorStore implementation?

william-silversmith:
Hi Stephen, 

I've been busy with [Kimimaro](https://github.com/seung-lab/kimimaro) at the moment, so I haven't yet fully digested the new Sharding spec. It makes a lot of sense to me to shard skeletons and meshes to avoid the high file creation fees on Google and Amazon ($4 per million, but we make billions). I don't immediately understand the point of sharding segmentation though since in a way, it is already sharded.  

I've started planning on adding sharding for meshes and skeletons at least as those have been major cost centers for us. I was starting to invent my own sharding design when @nkemnitz told me there already was one on Thursday. (Thanks Jeremy!)

It's gonna bit at least a few weeks if not longer depending on my availability before sharding support is released though.

william-silversmith:
Skimming through the spec, I guess if you want to reduce the number of image chunks as well, that makes sense. Those have been the least of our problems fwiw though. I will note that not all servers that we use support HTTP `Range` queries, but I suppose so long as we don't use sharding it's fine.

jbms:
The new sharded format will be supported by Tensorstore but support in cloud-volume would certainly be useful as well.  The sharded format does have downsides as listed in the spec, but when using GCS, even aside from write costs it is also possible to write much faster and makes moving/copying the data much easier.

unidesigner:
@jbms Do you have a rough estimate on when Tensorstore will be released?

jbms:
Also, which server do you use that doesn't support HTTP Range queries?

william-silversmith:
We have some S3 emulation service set up, but while I was told to regard
operations as entire file, it also promises full emulation of the S3 api so
I should try it before I say conclusively that it doesn't. I figure Range
is a subtlety that is usually not asked about by most people.

[...]

PR #371, opened 2020-07-31
Image shard generation

nkemnitz:
Shard Write:
* fix: don't crop partially intersecting shards
* feat: Add progress flag to `make_shard`
* perf: Avoid repeated string concatenation

Shard Read:
* perf: Avoid copies when converting JPEG bytes to numpy array
  * 30ms -> 0.5ms per call
* perf: Bundle adjacent byte range requests
  * 1000 s -> 5 s for completely empty/black shard
  * 1 s -> 0.5 s for example in https://github.com/seung-lab/cloud-volume/issues/354

### DIFF
--- a/cloudvolume/chunks.py
+++ b/cloudvolume/chunks.py
@@ -151,7 +151,7 @@ def decode_npz(string):
 
 def decode_jpeg(bytestring, shape, dtype):
   img = Image.open(io.BytesIO(bytestring))
-  data = np.array(img.getdata(), dtype=dtype)
+  data = np.asarray(img, dtype=dtype).ravel()
 
   return data.reshape(shape, order='F')
 
--- a/cloudvolume/datasource/precomputed/image/__init__.py
+++ b/cloudvolume/datasource/precomputed/image/__init__.py
@@ -367,7 +367,7 @@ class PrecomputedImageSource(ImageSourceInterface):
         )
         pbar.update()
 
-  def make_shard(self, img, bbox, mip=None, spec=None):
+  def make_shard(self, img, bbox, mip=None, spec=None, progress=False):
     """
     Convert an image that represents a single complete shard 
     into a shard file.
@@ -390,26 +390,34 @@ class PrecomputedImageSource(ImageSourceInterface):
       else:
         raise ValueError("mip {} does not have a sharding specification.".format(mip))
 
+    bbox = Bbox.create(bbox)
     if bbox.subvoxel():
       raise ValueError("Bounding box is too small to make a shard. Got: {}".format(bbox))
 
-    bbox = Bbox.create(bbox, self.meta.bounds(mip), bounded=True)
+    # Alignment Checks:
+    # 1. Aligned to atomic chunks - required for grid point generation
     aligned_bbox = bbox.expand_to_chunk_size(
       self.meta.chunk_size(mip), offset=self.meta.voxel_offset(mip)
     )
-    aligned_bbox = Bbox.clamp(aligned_bbox, self.meta.bounds(mip))
-
     if bbox != aligned_bbox:
       raise exceptions.AlignmentError(
         "Unable to create shard from a non-chunk aligned bounding box. Requested: {}, Aligned: {}".format(
         bbox, aligned_bbox
       ))
 
+    # 2. Covers the dataset at least partially
+    aligned_bbox = Bbox.clamp(aligned_bbox, self.meta.bounds(mip))
+    if aligned_bbox.subvoxel():
+      raise exceptions.OutOfBoundsError("Shard completely outside dataset: Requested: {}, Dataset: {}".format(
+        bbox, self.meta.bounds(mip)
+      ))
+
     grid_size = self.grid_size(mip)
     chunk_size = self.meta.chunk_size(mip)
     reader = sharding.ShardReader(self.meta, self.cache, spec)
 
-    gpts = lambda: gridpoints(bbox, self.meta.bounds(mip), chunk_size)
+    # 3. Gridpoints all within this one shard
+    gpts = lambda: gridpoints(aligned_bbox, self.meta.bounds(mip), chunk_size)
     all_same_shard = bool(reduce(lambda a,b: operator.eq(a,b) and a,
       map(reader.get_filename,
         map(lambda gpt: compressed_morton_code(gpt, grid_size), gpts())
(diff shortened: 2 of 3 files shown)


cloudvolume/datasource/precomputed/image/__init__.py
```diff
@@ -392,25 +392,39 @@ def make_shard(self, img, bbox, mip=None, spec=None):
...
-
     if bbox != aligned_bbox:
       raise exceptions.AlignmentError(
         "Unable to create shard from a non-chunk aligned bounding box. Requested: {}, Aligned: {}".format(
         bbox, aligned_bbox
       ))
 
+    # 2. Covers the dataset at least partially
+    aligned_bbox = Bbox.clamp(aligned_bbox, self.meta.bounds(mip))
+    if any(aligned_bbox.size3()) is False:
```
william-silversmith, line 412: I'd suggest `aligned_bbox.subvoxel()` to cover cases where it's less than one voxel but more than zero.

cloudvolume/datasource/precomputed/sharding.py
```diff
@@ -740,10 +739,11 @@ def synthesize_shard_file(spec, label_group, progress=False, presorted=False):
       minishard_index[0, i] = label - last_label
       minishard_index[1, i] = 0 # minishard_index[2, i - 1]
       minishard_index[2, i] = len(binary)
-      minishard += binary
+      minishard_components.append(binary)
       last_label = label
       del minishardgrp[label]
-    
+
+    minishard = b"".join(minishard_components)
```
william-silversmith, line 746: This is cool, but is this validated by performance tests?

nkemnitz: Yeah, that one is/was the biggest bottleneck:

With `x = b'a' * 1000`
```python
res = b''
for i in range(1000):
  res += x
```
`34.6 ms ± 505 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)`  - which makes sense; strings are immutable, so you end up allocating/discarding a new copy every time.

```python
tmp = []
for i in range(1000):
  tmp.append(x)
res = b''.join(tmp)
```
`147 µs ± 1.34 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)`

Pretty sure it's related to https://stackoverflow.com/questions/1349311/python-string-join-is-faster-than-but-whats-wrong-here/1350289#1350289
And I can completely understand his rant, because if you do that same experiment above with `str` instead of `bytes`, it appears to take that magical optimization path he is talking about and you suddenly end up with similar performance for both methods (with `join` still slightly faster, though)

Using list comprehension would be slightly faster again, but with the change above, most of the time is now spent copying and compressing the minishard components

william-silversmith: How long in wall clock time was make shard taking you then and now? Thanks for the info, didn't know that about strings. I think I take the side of the optimizers as real newbies aren't going to be researching the right way to do it. In this case, I guess I'm not a newbie, but I just figured I'd live with whatever performance += would give me until it became a crisis. Thanks for looking into it!

nkemnitz: Don't have the exact numbers at hand. Can check once the cluster is free again. `make_shard` went from approximately 2000 s to 500 s for my 8192x8192x512 vx shards (raw+gzip). Most of that is spent on gzipping each of the 32,768 tiny 256x256x16 components.

Another ~60 s is numpy [copying those cutouts into contiguous arrays](https://github.com/seung-lab/cloud-volume/blob/master/cloudvolume/chunks.py#L137)

william-silversmith: That's a big volume.... those numbers sound reasonable.

william-silversmith: I think certain brotli compression levels are faster than gzip for approximately the same compression. CloudFiles now supports brotli for all supported providers.

nkemnitz: That's correct, although currently not supported in Neuroglancer - for sharded volumes, decoding is done by the [async_computation workers](https://github.com/seung-lab/neuroglancer/tree/master/src/neuroglancer/async_computation).

william-silversmith:
https://github.com/seung-lab/cloud-volume/issues/264 is related to eeb55a2

william-silversmith:
If downloading entire shards, I really recommend just downloading the file and using `reader.disassemble_shard`.

cloudvolume/chunks.py
```diff
@@ -151,7 +151,7 @@ def decode_npz(string):
 
 def decode_jpeg(bytestring, shape, dtype):
   img = Image.open(io.BytesIO(bytestring))
-  data = np.array(img.getdata(), dtype=dtype)
+  data = np.asarray(img, dtype=dtype).ravel()
```
william-silversmith, line 154: should be `ravel(order='F')`

william-silversmith: Default order is C: https://numpy.org/doc/stable/reference/generated/numpy.ravel.html

nkemnitz:
> If downloading entire shards, I really recommend just downloading the file and using `reader.disassemble_shard`.

You mean I should check if bounding box envelops an entire shard and have a special code path to optimize for that? Or that the user should be responsible for downloading and disassembling shards?

william-silversmith:
You might want to run a test to see if there's a significant difference in
speed. One advantage of downloading the whole file is I can compute the md5
so that's nice.

cloudvolume/datasource/precomputed/image/__init__.py
```diff
@@ -390,26 +390,34 @@ def make_shard(self, img, bbox, mip=None, spec=None):
       else:
         raise ValueError("mip {} does not have a sharding specification.".format(mip))
 
+    bbox = Bbox.create(bbox)
```
william-silversmith, line 393: This change means you can't use slices with an unspecified end. Bbox.create(slices, bounds, bounded=True) calls reify_slices.

cloudvolume/datasource/precomputed/sharding.py
```diff
@@ -586,18 +589,42 @@ def get_data(self, label, path="", progress=None):
       for ext in exists.values()
     )
 
+    # Requesting many individual shard chunks is slow, but due to z-ordering
+    # we might be able to combine adjacent byte ranges. Especially helpful
+    # when downloading entire shards!
+    bundles = []
+    for chunk in sorted(files, key=itemgetter("path", "start")):
```
william-silversmith, line 596: `sorted` creates a new array vs `.sort` does it in-place. Would `.sort` be faster? Is it substantial compared to the rest?

cloudvolume/datasource/precomputed/image/__init__.py
william-silversmith: On the other hand this is probably a useless feature here.

william-silversmith:
LGTM

spikeliu:
Combined with the information from #221, I am little confused now, does cloud-volume support generating a volume in sharded format and upload it, or it only supports shard reads? I looked at the make_shard method, and it seems to generates a sharded file in memory, so at least it limits how large the file can be (can not beyond machine's memory). Also, it looks like there is no flag or method exposed to users for specifying sharded format when uploading the volume.

william-silversmith:
This is kind of a dumb state of affairs, but we currently have the ability to create a sharded volume from an unsharded one easily using Igneous, but we don't have an easy way to create one from scratch with CloudVolume other than via `make_shard` + CloudFiles upload. I haven't fully thought it through how one might do that, but the main reason is that creating new datasets from e.g. tiffs is fairly idiosyncratic. The shard specification dictates how big an uploaded image shard is supposed to be. Igneous now can generate a reasonable sharding specification automatically, so perhaps we could just tell the user what size image blocks to use...

william-silversmith:
Here's how we generate new sharded volumes:

https://github.com/seung-lab/igneous/blob/master/igneous/tasks/image.py#L472-L531
https://github.com/seung-lab/igneous/blob/master/igneous/task_creation/image.py#L387-L488

PR #375, opened 2020-08-05
feat: create secrets argument

### LINKED ISSUE #311: Auth tokens should be passable to Graphene
The auth tokens that we are converging on for pychunkgraph and other endpoints are not limited to cloudvolume, and it's even conceivable that someone might need them to programmatically access data even without cloudvolume installed. I can also see cases where, for testing, we need to use different tokens with different access levels. It would thus be good to not purely hard code the tokens to a cloudvolume secrets file, but also allow a graphene cloudvolume to accept an auth token as an argument on initialization.

william-silversmith:
Enables using CloudVolume like so:

```python
cv = CloudVolume(..., secrets={ ... })
```

Resolves #311


william-silversmith:
For graphene... what if we want to provide a GCS token and a graphene token???

william-silversmith:
For most use cases, that won't be a problem because you can use `use_https=True` to read the GCS volume. Maybe this is a perfect v enemy of the good situation.

william-silversmith:
@sdorkenw Graphene always requires a token these days right? It's safe to check for no token and error if none provided?

william-silversmith:
Sven confirmed that it always needs a token nowadays.

PR #388, opened 2020-08-28
fix: ensure segid assigned to meshes

william-silversmith:


fcollman:
This appears to fix some of the errors, but is now not properly connecting all the mesh fragments that exist.

william-silversmith:
We're continuing to discuss this in #389

PR #389, opened 2020-08-28
changing logic for speculative manifests

fcollman:
I think this change uses the existing non-speculative code path, but inserts conditional logic to download initial meshes using the precomputed sharded interface, so that that code can compute the byte offsets from the headers rather than relying upon the server to do so, while still downloading dynamic meshes using the old precomputed non-sharded interface.

### DIFF
--- a/cloudvolume/datasource/graphene/mesh/sharded.py
+++ b/cloudvolume/datasource/graphene/mesh/sharded.py
@@ -112,7 +112,7 @@ class GrapheneShardedMeshSource(GrapheneUnshardedMeshSource):
     dynamic_labels.update(initial_labels)
     return dynamic_labels
 
-  def parse_manifest_filenames(self, manifest):
+  def parse_manifest_filenames(self, manifest, verify=True):
     lists = defaultdict(list)
     initial_regexp = re.compile(r'~(\d+)/([\d\-]+\.shard):(\d+):(\d+)')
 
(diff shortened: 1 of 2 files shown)


cloudvolume/datasource/graphene/mesh/sharded.py
```diff
@@ -161,30 +165,33 @@ def get_meshes_via_manifest_byte_offsets(self, seg_id, bounding_box):
...
-    initial_meshes = []
-    while files:
-      f = files.pop()
-      mesh = Mesh.from_draco(f['content'])
-      start, end = f['byte_range']
-      key = (f['path'], start, end)
-      mesh.segid = segid_map[key]
-      initial_meshes.append(mesh)    
-
+    if verify:
```
william-silversmith, line 168: Just as a style point, it's usually a not-so-good idea to put long blocks of code under an if/else because it forces the reader to remember which level of indentation they're in. For example, in this case it might be more readable to write:

```python
if not verify:
      segid_map = self.get_meshes_on_bypass(lists['initial'], allow_missing=True)
      initial_meshes = [ v for v in segid_map.values() ]
      return dynamic_meshes + initial_meshes

# verify code follows
```

cloudvolume/datasource/graphene/mesh/sharded.py
```diff
@@ -292,7 +299,7 @@ def download_segid(self, seg_id, bounding_box, bypass=False, use_byte_offsets=Fa
     if bypass:
       mesh = self.get_meshes_on_bypass(seg_id)[seg_id]
     else:
-      meshes = self.get_meshes_via_manifest(seg_id, bounding_box, use_byte_offsets=use_byte_offsets)
+      meshes = self.get_meshes_via_manifest_byte_offsets(seg_id, bounding_box, verify=use_byte_offsets)
```
william-silversmith, line 302: Ah, I see there was no `use_byte_offsets` defined in this function.

william-silversmith:
I'm not sure how this is substantively different from `get_meshes_via_manifest_labels` which checks dynamic meshes then initial meshes without relying on the server for more than getting which labels to examine from the graph.

```
  def get_meshes_via_manifest_labels(self, seg_id, bounding_box):
    level = self.meta.meta.decode_layer_id(seg_id)
    labels = self.get_fragment_labels(seg_id, level=level, bbox=bounding_box)
    meshes = self.get_meshes_on_bypass(labels, allow_missing=True) # sometimes a tiny label won't get meshed
    return list(meshes.values())

  def get_fragment_labels(self, segid, lod=0, level=2, bbox=None, bypass=False):
    if bypass:
      return [ segid ]

    manifest = self.fetch_manifest(segid, lod, level, bbox, return_segids=True, verify=False)
    return manifest["seg_ids"]
```

william-silversmith:
I see, the main difference is using the dynamic mesh filename provided by the server vs computing one. It's possible I've made an error in the `compute_filename` code as it was written when there were no dynamic meshes.

william-silversmith:
I got compute_filename working thanks to @manuel-castro and it seems to work like a charm, so I'm closing this. We can revisit if we need additional performance via skipping dynamic checks.

PR #392, opened 2020-09-02
perf: use simplejpeg for JPEG encoding/decoding

william-silversmith:
simplejpeg uses an updated version of libjpeg-turbo which can be much faster than the version of libjpeg used by Pillow.

### DIFF
--- a/cloudvolume/chunks.py
+++ b/cloudvolume/chunks.py
@@ -90,7 +90,9 @@ def encode_jpeg(arr, quality=85):
     arr = arr[..., np.newaxis] # add channels to end of x,y,z
 
   num_channel = arr.shape[3]
-  reshaped = arr.reshape(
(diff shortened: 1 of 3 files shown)


william-silversmith:
As of 1.3.7, simplejpeg now has binaries for all major platforms. https://gitlab.com/jfolz/simplejpeg/-/issues/3

william-silversmith:
Nico wasn't kidding. In my limited testing, I am seeing ~4x faster encoding and 2-3x faster decoding.

william-silversmith:
Need to figure out if the way we're saving these is backwards compatible with older datasets.

image

william-silversmith:
Much better:

image

william-silversmith:
Tested with sharded jpegs on minnie too. Looks good!

PR #393, opened 2020-09-03
test(graphene): add tests for meshes based on real data

fcollman:
I have added some more tests to ensure that mesh access works properly with the newer graphene server protocols.

### DIFF
--- a/test/test_cv/sharded_info.json
+++ b/test/test_cv/sharded_info.json
@@ -1,0 +1,362 @@
+
+{
+    "app": {
+      "supported_api_versions": [
+        0,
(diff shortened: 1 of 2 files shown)


william-silversmith:
Looks like the use of f-strings is killing Python 3.5.
```
E     File "/home/travis/build/seung-lab/cloud-volume/test/test_graphene.py", line 94
E       f"manifest/{TEST_SEG_ID}:0?verify=True")
E   SyntaxError: invalid syntax
```

fcollman:
i have removed the fstring formatting line

PR #396, opened 2020-09-06
perf(sharding): compressed morton code and gridpoints

nkemnitz:
- [x] compressed_morton_code
- [ ] gridpoints / xyzrange

#### feat(compressed_morton_code): accept arrays as input
For shard generation and reading large cutouts, several thousand codes need to be calculated. By allowing the entire array as input, we properly utilize numpy:
```python
grid_size = Vec(520,440,260)
gpts = np.array([np.random.randint(520, size=65536), np.random.randint(440, size=65536), np.random.randint(260, size=65536)]).T

%timeit for gridpoint in gpts: compressed_morton_code(gridpoint, grid_size)
# 4.2 s ± 168 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%timeit compressed_morton_code(gpts, grid_size)
# 8.16 ms ± 350 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

### DIFF
--- a/cloudvolume/datasource/precomputed/image/common.py
+++ b/cloudvolume/datasource/precomputed/image/common.py
@@ -68,18 +68,26 @@ def gridpoints(bbox, volume_bbox, chunk_size):
     yield Vec(x,y,z)
 
 def compressed_morton_code(gridpt, grid_size):
+  gridpt = np.asarray(gridpt, dtype=np.uint32)
+  single_input = False
+  if gridpt.ndim == 1:
+    gridpt = np.atleast_2d(gridpt)
+    single_input = True
+
   code = np.uint64(0)
-  num_bits = int(max([ np.ceil(np.log2(size)) for size in grid_size ]))
+  num_bits = max(( math.ceil(math.log2(size)) for size in grid_size ))
   j = np.uint64(0)
   one = np.uint64(1)
 
   for i in range(num_bits):
     for dim in range(3):
       if 2 ** i <= grid_size[dim]:
-        bit = (((np.uint64(gridpt[dim]) >> np.uint64(i)) & one) << j)
+        bit = (((np.uint64(gridpt[:, dim]) >> np.uint64(i)) & one) << j)
         code |= bit
         j += one
-  
+
+  if single_input:
+    return code[0]
   return code
 
 def shade(dest_img, dest_bbox, src_img, src_bbox):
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by william-silversmith** [comment]
Do you want to break this into two PRs so we can get the first speedup merged? This is v. cool.

**Comment by nkemnitz** [comment]
Feel free to merge.

File: cloudvolume/datasource/precomputed/image/common.py [hunk 1]
```diff
@@ -68,18 +68,26 @@ def gridpoints(bbox, volume_bbox, chunk_size):
     yield Vec(x,y,z)
 
 def compressed_morton_code(gridpt, grid_size):
+  gridpt = np.asarray(gridpt, dtype=np.uint32)
```
william-silversmith, line 71: This is fine, but you might want to know about `lib.toiter(gridpt, return_multple=True)`

cloudvolume/datasource/precomputed/image/common.py
```diff
@@ -68,18 +68,26 @@ def gridpoints(bbox, volume_bbox, chunk_size):
...
+  single_input = False
+  if gridpt.ndim == 1:
+    gridpt = np.atleast_2d(gridpt)
+    single_input = True
+
   code = np.uint64(0)
```
william-silversmith, line 77: Now that code is vectorized, the `|=` operator below appears magic. I had to do some experimenting to understand what happens:

```python
>>> import numpy as np
>>> code = np.uint64(0)
>>> bit = np.array([[1,2,3],[4,5,6]], dtype=np.uint64)
>>> bit
array([[1, 2, 3],
       [4, 5, 6]], dtype=uint64)
>>> bit >> np.uint64(1)
array([[0, 1, 1],
       [2, 2, 3]], dtype=uint64)
>>> bit >> np.uint64(1) << np.uint64(1)
array([[0, 2, 2],
       [4, 4, 6]], dtype=uint64)
>>> code | (bit >> np.uint64(1) << np.uint64(1))
array([[0, 2, 2],
       [4, 4, 6]], dtype=uint64)
>>> code | (bit[:,0] >> np.uint64(1) << np.uint64(1))
array([0, 4], dtype=uint64)
```

william-silversmith: This could be fixed by assigning code to be an array from the start instead of a scalar.

cloudvolume/datasource/precomputed/image/common.py
```diff
@@ -68,18 +68,26 @@ def gridpoints(bbox, volume_bbox, chunk_size):
     yield Vec(x,y,z)
 
 def compressed_morton_code(gridpt, grid_size):
+  gridpt = np.asarray(gridpt, dtype=np.uint32)
+  single_input = False
+  if gridpt.ndim == 1:
+    gridpt = np.atleast_2d(gridpt)
```
william-silversmith, line 74: I had no idea this existed. I love it.

PR #399, opened 2020-09-09
fix(windows,paths): non-file windows parsing was broken

william-silversmith:
Windows parsing was dropping the bucket name from the path.


william-silversmith:
Thanks to Christa for reporting!

PR #403, opened 2020-09-10
fixing level=2 for bounding box queries

fcollman:
When doing a bounding box query with graphene, it doesn't make sense to specify something other than level2 because you don't want to get stitched fragments that extend well beyond your bounding box in some inconsistent fashion depending on the state of mesh stitching in the backend. 

This simply sets the level=2 when specifying a bounding box. 

I have a corresponding PR on the PCG backend to make this functional. 
https://github.com/seung-lab/PyChunkedGraph/pull/229

### DIFF
--- a/cloudvolume/datasource/graphene/mesh/sharded.py
+++ b/cloudvolume/datasource/graphene/mesh/sharded.py
@@ -146,7 +146,8 @@ class GrapheneShardedMeshSource(GrapheneUnshardedMeshSource):
     """
     level = self.meta.meta.decode_layer_id(seg_id)
     dynamic_cloudpath = self.meta.join(self.meta.meta.cloudpath, self.dynamic_path())
-
+    if bounding_box is not None:
+      level=2
     manifest = self.fetch_manifest(seg_id, level=level, bbox=bounding_box, return_segids=True)
     lists = self.parse_manifest_filenames(manifest)
 
@@ -189,6 +190,8 @@ class GrapheneShardedMeshSource(GrapheneUnshardedMeshSource):
 
   def get_meshes_via_manifest_labels(self, seg_id, bounding_box):
     level = self.meta.meta.decode_layer_id(seg_id)
+    if bounding_box is not None:
+      level=2
     labels = self.get_fragment_labels(seg_id, level=level, bbox=bounding_box)
     meshes = self.get_meshes_on_bypass(labels, allow_missing=True) # sometimes a tiny label won't get meshed
     return list(meshes.values())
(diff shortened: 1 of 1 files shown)


fcollman:
akhilesh has now made this change part of 
https://github.com/seung-lab/PyChunkedGraph/pull/234#pullrequestreview-488752901

but when we deploy this change it will break compatibility with the current cloud-volume master.  This is because cloud-volume is incorrectly asking for speculative meshes starting at the level of the nodeID.  Only the server knows what the maximum stitch level for a dataset is, so it should not pass anything, unless it wants to specify that meshes are very low level meshes (level=2), such as in the case that the desired behavior is to have a bounding box.

cloudvolume/datasource/graphene/mesh/sharded.py
```diff
@@ -188,7 +189,9 @@ def get_meshes_via_manifest_byte_offsets(self, seg_id, bounding_box):
     return dynamic_meshes + initial_meshes
 
   def get_meshes_via_manifest_labels(self, seg_id, bounding_box):
-    level = self.meta.meta.decode_layer_id(seg_id)
+    level = None
```
william-silversmith, line 192: I don't think I understand what this is doing here. What levels are specified if there is no bounding box? Does the server decide arbitrarily? Was it wrong to specify the level beforehand?

fcollman: The levels parameter is an optional one in the api, it means that the PCG should only return nodeIDs starting at this level in the oct-tree hierarchy.  The old code was saying that it should start with fragments that start at the level of the requested ID.  The problem is that the meshing might only be stitched up to a certain level, like level 7 for minnie. So when you combine that with speculative meshes, you are telling the api that you don't want it to check if a mesh exists and you want meshes at the level of the seg_id you are requesting (which for root_id's will be level 12), and you don't want it to check whether the mesh exists, just send you the path based upon the timestamp of the node.  

The old implementation was actually ignoring the level parameter if it was passed via speculative meshes.  This breaks a different function, when you actually do want to specify level=2 when you are doing bounding box queries because you don't want it returning fragments from higher up the hierarchy as the bounding box calculation is going to be messy in such a case.

fcollman: the new implementation fixes this and pays attention to the level parameter, but now the old cloud-volume code causes a bug where it returns a pointer to the level 12 nodeID, which then cloud volume discovers does not exist, silently ignores that a fragment doesn't exist, and then errors because its list of meshes is empty.  

The server is the entity that knows what level meshes are stitched up to, so really not passing level is the correct thing to do for speculative meshes.

william-silversmith: Gottcha. Sorry for taking so long to get to this. That one line was causing me anxiety and I should have taken the time to ask before.

PR #422, opened 2020-10-28
perf: faster decode of spatial_index json

william-silversmith:
It looks like on this test, the decode time was faster by about 30%, but the cost is a huge increase in memory usage. It also creates duplicated code which will be expensive to maintain. The real situation requires > 100k files whereas this test was on about 1200 files, so the memory problem will be 100x worse.

image


william-silversmith:
Using pysimdjson features ended up being better on every dimension.

PR #427, opened 2020-11-13
fix(secrets): warning about default application credentials environme…

### LINKED ISSUE #424: Default gcloud permissions not successfully used when secret.json is missing?
When there's no file at `~/.cloudvolume/secrets/google-secret.json`, cloudvolume claims that it will use the "default Google credentials". https://github.com/seung-lab/cloud-volume/blob/master/cloudvolume/secrets.py#L89

I would expect this to mean that if I've authenticated using `gcloud init` or `gcloud auth login` that cloudvolume would be able to read and write from any buckets that `gsutil` commands could read and write from. That doesn't seem to be the case for me – when I don't have a `google-secret.json` file, commands like `vol = CloudVolume(cloud_path, info=info)` just print the line of yellow text from the link above and then python hangs. (In contrast, I can use `gsutil` commands to interact with the same bucket.) Am I doing something wrong, am I misunderstanding the meaning of "default Google credentials", or is this a bug?

william-silversmith:
…nt var

Resolves #424

@jasper-tms what do you think?

### DIFF
--- a/cloudvolume/secrets.py
+++ b/cloudvolume/secrets.py
@@ -87,6 +87,16 @@ def google_credentials(bucket = ''):
 
   if google_credentials == None:
     print(colorize('yellow', 'Using default Google credentials. There is no ~/.cloudvolume/secrets/google-secret.json set.'))  
+    if os.environ.get("GOOGLE_APPLICATION_CREDENTIALS", None) is None:
+      print(colorize('yellow', 
+      """
+      Warning: Environment variable GOOGLE_APPLICATION_CREDENTIALS is not set.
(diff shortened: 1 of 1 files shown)


william-silversmith:
Actually, the action takes place in cloud-files now. I'll move this change there.

jasper-tms:
Yes, this looks good. I pulled the updated cloud-files and tried it out, and yea as desired this warning message and help text printed just before the connection attempt started. I imagine this will be good enough for most users to get to a point where they can use their default login credentials and so won't have to tinker with service accounts. Thanks Will.

PR #433, opened 2020-12-15
make functions without parameter as propterty

jingpengw:
I kind of likes property than function.
this is a breaking change though...


william-silversmith:
I'm not sure this would have a good cost-benefit tradeoff for the ecosystem. The automated tests are failing internally. Igneous would be another piece of software that would have to be checked too. I'm not sure how people are using these in the wild too...

I do agree that often these are prettier as properties, but I also like to signify that a computation is happening vs a simple read via the parens.

jingpengw:
might not worth changing it, I'll close it now.

PR #434, opened 2020-12-19
perf: support parallel file fetching for cache, sharded, and mesh bypass

william-silversmith:


ceesem:
Don't merge this just yet! I am getting a segfault when doing a segmentation download with `cv.download(bbox, seg_ids=seg_ids)` when using a cloudvolume instance initialized with a `parallel` argument higher than 1. It looks like multithreading is doubled up in that instance. It doesn't happen when seg_ids is not set.

william-silversmith:
That's the kind of thing I was worried about. Thanks Casey! I'll look into it.

william-silversmith:
I managed to reproduce the segmentation fault on my machine reliably. Good first step.

```from cloudvolume import CloudVolume, Bbox
import numpy as np

if __name__ == "__main__":
	cv = CloudVolume(graphene minnie, 
	  progress=True, parallel=2, use_https=True)
	pt = [250953, 237168, 20451] // np.array([2,2,1])
	bbox = Bbox(pt, pt-[100, 100, 2])
	cv.download(bbox, segids=[864691135696000154]) # this must occur before mesh get
	cv.mesh.get(864691135696000154)
```

william-silversmith:
The segfault is caused when examining the image after passing through `fastremap.remap` with `in_place=True` and also assigning `img = fastremap.remap(..., in_place=True)`. Removing the assignment seems to fix it.

william-silversmith:
Ah, I think I know what's happening. `img = ...` is causing the original reference to VolumeCutout to be garbage collected, while the new image is a stride tricked reference to the underlying buffer. Once the underlying buffer disappears, you get a segfault.

I was able to cure the segfault by also writing `img2 = img`.

william-silversmith:
I pushed the fix (700a7e67a130c57f4366a030f79427b3442d4abe) to master (by accident) but the issue was present there too, so that shouldn't stop this PR from going through.

william-silversmith:
@ceesem any remaining issues with this PR? The segfault you noticed (THANK YOU) is fixed.

ceesem:
I had no issues in other contexts, and it'll be a useful feature.

PR #438, opened 2021-01-26
fix: background_color not applied when bounded=False, fill_missing=True

macrintr:
Currently, only works for parallel=1

### DIFF
--- a/cloudvolume/chunks.py
+++ b/cloudvolume/chunks.py
@@ -60,12 +60,14 @@ def encode(img_chunk, encoding, block_size=None):
   else:
     raise NotImplementedError(encoding)
 
-def decode(filedata, encoding, shape=None, dtype=None, block_size=None):
+def decode(filedata, encoding, shape=None, dtype=None, block_size=None, 
+                     background_color=0):
   if (shape is None or dtype is None) and encoding not in ('npz', 'fpzip', 'kempressed'):
     raise ValueError("Only npz encoding can omit shape and dtype arguments. {}".format(encoding))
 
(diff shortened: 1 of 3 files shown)


cloudvolume/chunks.py
```diff
@@ -60,12 +60,14 @@ def encode(img_chunk, encoding, block_size=None):
...
 
-def decode(filedata, encoding, shape=None, dtype=None, block_size=None):
+def decode(filedata, encoding, shape=None, dtype=None, block_size=None, 
+                     background_color=0):
   if (shape is None or dtype is None) and encoding not in ('npz', 'fpzip', 'kempressed'):
     raise ValueError("Only npz encoding can omit shape and dtype arguments. {}".format(encoding))
 
   if filedata is None or len(filedata) == 0:
-    return np.zeros(shape=shape, dtype=dtype)
+    print(background_color)
```
william-silversmith, line 69: Remove debug statement.

cloudvolume/chunks.py
```diff
@@ -60,12 +60,14 @@ def encode(img_chunk, encoding, block_size=None):
...
+    return np.full(shape=shape, fill_value=background_color, dtype=dtype)
```
william-silversmith, line 70: This rocks. I will start using `np.full` in more places.

william-silversmith (changes requested):
This PR rocks. Thank you so much Tommy! 

I think the parallel situation can be easily remedied by adding these two lines to `child_process_download`:

```python
  if background_color != 0:
    dest_img[dest_bbox.to_slices()] = background_color
```

PR #443, opened 2021-02-07
feat(requester pays): GCS and S3

### LINKED ISSUE #442: Feature request: Requester pays
https://github.com/seung-lab/cloud-files/issues/40 could be expanded to CloudVolume.

Will submit a PR shortly

nkemnitz:
Resolves #442 

Requires https://github.com/seung-lab/cloud-files/pull/41

### DIFF
--- a/cloudvolume/datasource/precomputed/image/__init__.py
+++ b/cloudvolume/datasource/precomputed/image/__init__.py
@@ -95,7 +95,7 @@ class PrecomputedImageSource(ImageSourceInterface):
     mip = mip if mip is not None else self.config.mip
     mip = self.meta.to_mip(mip)
     
-    cf = CloudFiles(self.meta.cloudpath, secrets=self.config.secrets)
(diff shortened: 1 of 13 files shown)


cloudvolume/datasource/precomputed/image/__init__.py
```diff
@@ -232,6 +233,8 @@ def upload(
       non_aligned_writes=self.non_aligned_writes,
       green=self.config.green,
       fill_missing=self.fill_missing, # applies only to unaligned writes
+      secrets=self.config.secrets,
```
nkemnitz, line 236: User defined secrets were not passed for uploads here

william-silversmith:
Need to update requirements for `cloud-files>=1.25.0`

william-silversmith:
I'm starting to think that we need a single object that we pass around everywhere that contains the global config options for CF. Manuel recently added the storage class parameter, I don't see why that won't eventually get added. It was fine when it was just passing secrets everywhere, but these options are multiplying.

PR #445, opened 2021-02-09
fix(encode_jpeg): correct 3D->2D shape for Neuroglancer

nkemnitz:
Array shape mixup in between the transposing / axis swapping / reshaping transformations.
Happened during the switch to simple_jpeg (#392)


william-silversmith:
Can you confirm with a visual inspection of a new upload? The code already there is likely correct as that's how I did it.

PR #447, opened 2021-02-13
fix(np.mmemap): tolerate during upload

william-silversmith:


william-silversmith:
Failing test was py37 failing to download due to a network error. py37 passed in the PR test, so it seems okay given Travis has somehow removed my button for rerunning tests.

PR #451, opened 2021-03-03
Float Resolution Support template

nkemnitz:
Hi Will - I saw your comment in #322.

@ranlu and I were hacking some floating point support together at Zetta a while ago. I wouldn't necessarily recommend to go this exact route (especially the changes to the lib.py might have unforeseen impact on other parts of cloud-volume and related packages), but it highlights some pitfalls with the floating point resolution we stepped into.


william-silversmith:
Handled in #467

PR #460, opened 2021-04-10
feat: add cave-secret.json and allow for fqdn qualification

### LINKED ISSUE #458: Support Multiple Graphene Server Tokens
We want to support multiple graph servers as there are now multiple copies in deployment.

The graphene team has also suggested that we revise `chunked-graph.json` to `cave.json`. 

CAVE: Connectome Annotation Versioning Engine  

We'll want to support a similar scheme to google and aws where we append a specifier to the file name. In the case of CAVE, we'll use the domain name.

```cave-some.server.com.json```

The dot structure is awkward, but what can you do. Replacing the dots with e.g. hyphens would mean another server could stand in for the original.

william-silversmith:
e.g. flywire.dynamicannotationframework.com-cave-secret.json
(fake example)

Also adds the ability to configure the `.cloudvolume` directory via an environment variable `CLOUD_VOLUME_DIR`

Resolves #458

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,16 +162,19 @@ You can create the `google-secret.json` file [here](https://console.cloud.google
 }
 ```
 
-#### `chunkedgraph-secret.json`
+#### `cave-secret.json`
 
-If you have a token from Graphene/Chunkedgraph server, create the `chunkedgraph-secret.json` file as shown in the example below.
-You may also pass the token to `GrapheneMetadata` class `auth_token="<your_token>"`, this will override the token from `json` file.
+*Note: used to be called chunkedgraph-secret.json. This is still supported but deprecated.*
+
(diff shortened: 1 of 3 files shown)


ceesem:
Hi Will,

These changes look extremely promising — I just wanted to mention that, since we use the same token for all of our services, it would be nice if you could give Sven/Forrest/me a heads-up before making a release with this PR so we can get a release of our FrameworkClient ready with the same logic integrated. It should be a quick change, but could cause a lot of confusion if things get out of sync.

william-silversmith:
This change should be backwards compatible so there shouldn't be a disruption I think?

william-silversmith:
@ceesem do you think releasing this would interfere with potential upgrades in the future?

ceesem:
I don't think so — we talked about this and decided that the right place for dealing with the endpoint logic is elsewhere. Please feel free to go ahead.

william-silversmith:
Thanks Casey!

Failing tests are not relevant. I should schedule some time to fix them.

Issue #450, opened 2021-02-28
mesh.get: vertex positions garbled


schlegelp:
Hi Will! 

I'm running into a strange error with the mesh download but I'm not sure it's an issue with cloudvolume directly.

I get different results when running the following code on my local machine (OSX) versus on [DeepNote](https://deepnote.com/) (that's a Jupyter notebook server running Linux):

```Python
# Download a random mesh from the Janelia hemibrain dataset 
vol = CloudVolume('precomputed://gs://neuroglancer-janelia-flyem-hemibrain/v1.1/segmentation', use_https=True)

mesh = vol.mesh.get(295478082, lod=3)[295478082] 
```

On my local machine the vertex positions are correct:

```Python
>>> mesh.vertices
array([[ 67711.03 , 111167.695, 131136.   ],
       [ 67711.03 , 111039.695, 131136.   ],
       [ 67647.03 , 111231.695, 131136.   ],
       ...,
       [ 66111.01 , 131136.   , 139008.12 ],
       [ 65855.   , 131136.   , 138752.11 ],
       [ 65983.01 , 131136.   , 138752.11 ]], dtype=float32)
```

On DeepNote they come out as nonsense:

```Python
>>> mesh.vertices
>>> array([[6.40000e+01, 6.40000e+01, 6.40000e+01],
           [6.40000e+01, 6.40000e+01, 6.40000e+01],
           [6.40000e+01, 6.40000e+01, 6.40000e+01],
           ...,
           [6.40000e+01, 1.31136e+05, 1.31136e+05],
           [6.40000e+01, 1.31136e+05, 1.31136e+05],
           [6.40000e+01, 1.31136e+05, 1.31136e+05]], dtype=float32)
```

On both, my local machine and DeepNote, I run `cloudvolume` `3.8.0` and DracoPy `0.0.15`. I'm aware that this will be rather tricky to debug but maybe you have a good idea?

william-silversmith:
This seems to pertain to the mutli-lod logic. @perlman would you have any insights? I'm planning on working more with the multi-lod format soon, but I haven't quite gotten there yet. Thanks!

william-silversmith:
Also, @schlegelp can you confirm whether or not the defect occurs when you run it outside of the notebook? Thanks!

nkemnitz:
Works on Windows + Python 3.8.8. Fails on the same machine with WSL (Ubuntu 20 + Python 3.8.5).

[...]

PR #465, opened 2021-04-23
Fix: multilod mesh vertex corruption

manuel-castro:
DracoPy <= 0.0.15 erroneously assumes that every vertex is a float32, by copying the draco vertex buffers into float arrays. This works with the current `multilod.py` code (in certain platforms) because the multilod code views the DracoPy float data as np.int32, so if the original vertices were also int32 the data is unchanged. DracoPy 0.0.19 fixes this earlier bug, by appropriately converting the vertices into floats if they are of a different type. Ironically that fix means that the multilod code is now broken, because it views the now converted float data as np.int32. Therefore, we should remove this np.view call. Talked with Will and we agreed that this is the appropriate fix, and that the view was likely originally added to get around DracoPy's bug. Additionally, I confirmed that DracoPy 0.0.19 without the view call returns the same vertices as 0.0.15 with it, and that google's draco decoder decodes the same mesh from the draco file as DracoPy 0.0.19.

The related issues are here: https://github.com/seung-lab/cloud-volume/issues/450 and https://github.com/seung-lab/DracoPy/issues/16

### DIFF
--- a/cloudvolume/datasource/precomputed/mesh/multilod.py
+++ b/cloudvolume/datasource/precomputed/mesh/multilod.py
@@ -127,11 +127,6 @@ class ShardedMultiLevelPrecomputedMeshSource(UnshardedLegacyPrecomputedMeshSourc
           continue
         mesh = Mesh.from_draco(frag_binary)
 
-        # Conversion references:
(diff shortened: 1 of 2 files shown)


william-silversmith (approved):

perlman:
Wow, I'm glad you found the root cause!

Thanks for the explanation.

PR #467, opened 2021-04-29
feat: Support for floating point resolution

william-silversmith:


william-silversmith:
Having difficulty figuring out how to reset the GOOGLE_SECRET... too many levels of bash escaping. Passes tests locally though.

PR #471, opened 2021-05-13
feat(get_roots): avoid sending zeros, dupes, and root ids to server

### LINKED ISSUE #469: Handling of zeros by .get_roots()
Passing lists of supervoxel IDs containing zeros to CloudVolume.get_roots() causes only the first zero in the list to be dropped, while subsequent zeros return a root ID of zero. I am not sure if this is intentional, but this could cause indexing headaches if it isn't noticed and handled afterwards.

william-silversmith:
Resolves #469

### DIFF
--- a/cloudvolume/frontends/graphene.py
+++ b/cloudvolume/frontends/graphene.py
@@ -379,6 +379,9 @@ class CloudVolumeGraphene(CloudVolumePrecomputed):
     return chunk_mappings
 
   def _get_roots_v1(self, segids, timestamp, binary=False, stop_layer=None):
+    if len(segids) == 0:
+      return []
+
     args = {}
 
     headers = {}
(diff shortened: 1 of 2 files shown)


jefferis:
Hi @william-silversmith just checking that we still get an array back? But you just dedup on the way to the server and then put everything back on the way back out to the user?

william-silversmith:
Yea, there should be no change to the user interface. I'm screening for zeros, duplicates, and IDs that are already root IDs. I send a unique array to the server and then remap it back to the input array. 

This seems inefficient to me (why don't we return a dict) but I don't want to break backwards compatibility.

jefferis:
> Yea, there should be no change to the user interface. I'm screening for zeros, duplicates, and IDs that are already root IDs. I send a unique array to the server and then remap it back to the input array.

Thanks for confirming. This is what I am already doing in my [own wrapper](https://github.com/natverse/fafbseg/blob/master/R/flywire-api.R#L148). I guess it will be useful for python people using cloudvolume directly. In my own experience this is actually the behaviour that the end user normally wants. For example you might have a dataframe with a set of xyz locations that you have mapped to supervoxels. You want to identify the rootid for each row and not worry about mapping a dictionary back onto the rows.

william-silversmith:
I suppose that's true, but these days it's pretty easy to call `fastremap.remap`. I just never like providing a basic API that's intentionally inefficient so that people with more pressing needs are locked in. :(

PR #482, opened 2021-06-26
fix: update morton coding

chrisroat:
For grid dimensions that are a power of 2, the code was attributing an extra bit.  Fixing that also seems to fix the prior problem of size=1 dimensions (a power of 2!) naturally, without the extra check.

I added a test for the case I recently hit, where the morton code was out of range (195, where the #chunks was 128).  In addition, a prior test was incorrect and updated.

### DIFF
--- a/cloudvolume/datasource/precomputed/image/common.py
+++ b/cloudvolume/datasource/precomputed/image/common.py
@@ -88,7 +88,7 @@ def compressed_morton_code(gridpt, grid_size):
 
   for i in range(max(num_bits)):
     for dim in range(3):
-      if 2 ** i <= grid_size[dim] and grid_size[dim] > 1:
+      if 2 ** i < grid_size[dim]:
         bit = (((np.uint64(gridpt[:, dim]) >> np.uint64(i)) & one) << j)
         code |= bit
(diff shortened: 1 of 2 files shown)


chrisroat:
@william-silversmith If you are creating a new igneous docker image, can you include this fix, as well?

william-silversmith:
Thanks Chris! I'll need to think about this, but I can say that if my code is wrong so is the spec. (!)
https://github.com/google/neuroglancer/blob/master/src/neuroglancer/datasource/precomputed/volume.md#compressed-morton-code

chrisroat:
No worries.  I just rolled my own igneous image.

Yeah, I had to think about it a while, too.  I was getting some nasty re-arragements of my data (shard-sized chunks in the wrong place, or entirely missing) when running downsampling, and this seems to fix them.

william-silversmith (approved):

william-silversmith:
I think you're right. This makes sense because a grid size of 8 means you need to be able to represent 0 through 7, but not 8.

william-silversmith:
I'll release a new Igneous once I make a new CV release.

PR #484, opened 2021-06-29
allowing translation for multi-lod-mesh

fcollman:


cloudvolume/datasource/precomputed/mesh/multilod.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, meta, cache, config, readonly=False):
     self.lod_scale_multiplier = self.meta.info['lod_scale_multiplier']
     self.transform = np.array(self.meta.info['transform'] + [0,0,0,1]).reshape(4,4)
   
-    if np.any(self.transform * np.array([[0,1,1,1],[1,0,1,1],[1,1,0,1],[1,1,1,0]])):
+    if np.any(self.transform * np.array([[0,1,1,0],[1,0,1,0],[1,1,0,0],[1,1,1,0]])):
       raise exceptions.MeshDecodeError(red("Non-scale homogeneous transforms are not implemented"))
```
william-silversmith, line 50: The error message is now incorrect. Translations are also supported.

cloudvolume/datasource/precomputed/mesh/multilod.py
```diff
@@ -170,7 +170,7 @@ def __init__(self, meta, cache, config, readonly=False):
...
```
william-silversmith, line 174: Same here.

william-silversmith (changes requested):
Thanks for the PR Forrest! Any reason not to use `matmul`? That would solve the general case and obviate the need for an error message.

fcollman:
@william-silversmith I've ditched the hack and now done the full matrix multiplication and tested it that it's working on the miinnie example i had with a translation component, and so have removed the assertions entirely now.

william-silversmith:
Thank you so much Forrest! I think this does solve the immediate problem, but I started looking through the code for where transform is defined. Here's what I found:

Sharded and Unsharded multi-res meshes define transform thusly: 
```python
self.transform = np.array(self.meta.info['transform'] + [0,0,0,1]).reshape(4,4)
```
The single res format does not even read `transform` from the info file. This was a feature that was added later. I should fix these problems.

There are two additional things I'd like to add to this PR:
1.  tests
2. skipping the matrix multiplication if the transform is the identity matrix

If that's too much extra work, I can try to take it on. I suspect if we don't do these things we'll start running into problems not too far into the future now that `transform` is appearing as non-identities.

fcollman:
I'm eyeing the public releases of minnie, so whatever route makes there be an easy to install version that can enable that short term, and then a path toward a "correct" version longer term is my preference.

fcollman:
i don't see the problem with what you've got there, assuming self.meta.info['transform'] is a 3x4 matrix where the last column is the translation component.   That loads that from the info file and nobody everyone has a 4x4 homogenous matrix.

Tests agreed, should be herer.

william-silversmith:
You're right. I missed that I was trying to make a homogeneous matrix. In that case, let's just address those last two points for now and I'll make an issue for upgrading the single resolution format.

fcollman:
so for tests you are currently using data in gs://seunglab-test/test_v0/segmentation for some things, I added test files for graphene in the directory.. which model do you want to use? I can't add test data to the bucket, but we need a cloudvolume with some mesh files and a transform in the info to verify that things are working as expected.

william-silversmith:
There's also data located in https://github.com/seung-lab/cloud-volume/tree/master/test/test_cv If the data are not large, it's better to put it in there as the remote are a roadblock for others to run the unit tests. I can adjust the remote data though if necessary.

william-silversmith:
I added support for single resolution to this PR and added a fast path for the identity matrix. One test does *kind of* already cover the transform path but doesn't check to see if the vertices are sensible. I visually inspected the output and it seemed reasonable.

william-silversmith:
Okay, I added a few basic tests. I'll merge and release this tomorrow.

fcollman:
thank you will!

william-silversmith (approved):

PR #485, opened 2021-06-30
fix: raise errors from cloud files download

william-silversmith:
rx was not checking the cloud files error.

This raise is not behind a retry, so it should kill the process.

### DIFF
--- a/cloudvolume/datasource/precomputed/image/rx.py
+++ b/cloudvolume/datasource/precomputed/image/rx.py
@@ -293,6 +293,9 @@ def download_chunk(
   (file,) = CloudFiles(cloudpath, secrets=secrets).get([ filename ], raw=True)
   content = file['content']
 
+  if file['error']:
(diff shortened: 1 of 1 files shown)


nkemnitz (commented):
That should work.
Are there other storage formats / data types that are doing something similar?

william-silversmith:
At least for that example, we're safe. It called `cv.cache.download` which has the proper checks.

https://github.com/seung-lab/cloud-volume/blob/7b13d70c823e0af4987b13f0b2086fc337029b04/cloudvolume/cacheservice.py#L477-L487 

Really, I should have a flag for CloudFiles that automatically raises errors. It should be enabled by default.

william-silversmith:
I worked something up here: https://github.com/seung-lab/cloud-files/pull/57

william-silversmith:
This is covered by the update of cloud-files to 2.0.0 which raises errors by default.

PR #487, opened 2021-07-02
fix: make parallel image IO more robust

william-silversmith:
- Consolidates progress bar
- Checks for errors at end
- Fixes segfault on MacOS
- Accepts iterators
- Fixes file lock synchronization issue on MacOS

### DIFF
--- a/cloudvolume/datasource/precomputed/image/__init__.py
+++ b/cloudvolume/datasource/precomputed/image/__init__.py
@@ -23,7 +23,7 @@ from .... import sharedmemory, chunks
 
 from ... import autocropfn, readonlyguard, ImageSourceInterface
 from .. import sharding
-from .common import chunknames, gridpoints, compressed_morton_code, init_fs_lock
+from .common import chunknames, gridpoints, compressed_morton_code
 from . import tx, rx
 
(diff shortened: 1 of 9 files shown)


william-silversmith:
The new parallel implementation is hanging more than the old one. Could it be because it's more efficient and triggering rate limiters on GCS? Is it using too many open network connections?

It is generating a lot of these errors:
```
HTTPSConnectionPool(host='oauth2.googleapis.com', port=443): Max retries exceeded with url: /token (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x1c5baebb0>: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known'))
```

william-silversmith:
This problem doesn't appear when running CV in parallel using parallel=1 like so:
```bash
python test3.py&
python test3.py&
python test3.py&
wait
```

I wonder if the fork/exec is probably doing something weird to the Google Cloud Python library. After all, I had to call `reset_connection_pool` explicitly for my own code.

william-silversmith:
It would be nice if we can also improve the performance as well, but this is a pretty good improvement on its own.

PR #488, opened 2021-07-10
feat(Bbox): allow negative and tuple or array amt to grow bounding box

jingpengw:


cloudvolume/lib.py
```diff
@@ -633,8 +634,11 @@ def volume(self):
...
     return (self.minpt + self.maxpt) / 2.0
 
-  def grow(self, amt):
-    assert amt >= 0
+  def grow(self, amt: Union[int, tuple, list]):
+    if isinstance(amt, tuple) or isinstance(amt, list):
+      assert len(amt) == 3
+      amt = np.asarray(amt)
+
     self.minpt -= amt
```
william-silversmith, line 642: It looks like `shrink` has a check for valid Bbox that `grow` doesn't. Should this be incorporated?

william-silversmith: Should shrink take negative factors too?

jingpengw: Oh, I did not notice that there is a shrink function!
I just added another function called adjust. please take a look.

william-silversmith: This seems good, but shouldn't shrink be the negative of grow?

jingpengw: good point! did not realize that it should be negative!

jingpengw: I have corrected this in the following commit, please take a look.

william-silversmith:
Okay LGTM!

william-silversmith:
Thank you Jingpeng!

PR #503, opened 2021-10-16
BREAKING: refactor(MultiLevelPrecomputedMeshManifest): add writing

william-silversmith:
Can now read and write the binary format.

The breaking change is the change to the `cloudvolume.datasource.precomputed.mesh.multilod.MultiLevelPrecomputedMeshManifest` interface, but that should be used nearly exclusively internally.

- Fixes bug in spatial_index
- Adds mysql support for spatial_index


william-silversmith:
This is getting too big let's merge it and continue in another PR.

PR #508, opened 2021-11-15
make contain more accurate for boundary points

jingpengw:
following the convention of python indexing, the `maxpt` should not be inclusive?
according to the measurement of `size` or `length`, it seems that this is true.
if so, the point should not be contained if it is on the coordinate of `maxpt`?

### DIFF
--- a/cloudvolume/lib.py
+++ b/cloudvolume/lib.py
@@ -476,15 +476,9 @@ class Bbox(object):
       ]
     """
     arr = np.array(arr, dtype=np.float32)
-
-    mins = []
-    maxes = []
-
-    for i in range(arr.shape[1]):
-      mins.append( np.min(arr[:,i]) )
-      # Note that the right side should be exclusive
-      # following the python indexing scheme
-      maxes.append( np.max(arr[:,i]) + 1)
+    
+    mins = np.min(arr, axis=0)
+    maxes = np.max(arr, axis=0) + 1
 
     return Bbox( mins, maxes, dtype=np.int64)
 
--- a/cloudvolume/lib.py
+++ b/cloudvolume/lib.py
@@ -482,7 +482,9 @@ class Bbox(object):
 
     for i in range(arr.shape[1]):
       mins.append( np.min(arr[:,i]) )
-      maxes.append( np.max(arr[:,i]) )
+      # Note that the right side should be exclusive
+      # following the python indexing scheme
+      maxes.append( np.max(arr[:,i]) + 1)
 
     return Bbox( mins, maxes, dtype=np.int64)
 
--- a/cloudvolume/lib.py
+++ b/cloudvolume/lib.py
@@ -761,7 +761,7 @@ class Bbox(object):
 
     Returns: boolean
     """
-    return np.all(point >= self.minpt) and np.all(point <= self.maxpt)
+    return np.all(point >= self.minpt) and np.all(point < self.maxpt)
 
   def contains_bbox(self, bbox):
     return self.contains(bbox.minpt) and self.contains(bbox.maxpt)


jingpengw:
In that case, we should also change the `from_points` function?
the `maxpt` should `+1`

william-silversmith:
You have a good point. I think the dominant use of the bbox is the inclusive-exclusive formulation, and it's not so risky to change those two functions. Would you like to patch `from_points` as well?

jingpengw:
@william-silversmith I have added the `from_points` function. please take a look.

william-silversmith:
Thank you Jingpeng! One small item. I think `from_points` could be written more efficiently. Would you mind replacing the loop with `np.min(arr, axis=0)` and `np.max(arr, axis=0) + 1`? Not your fault, but a good opportunity to make things work better.

jingpengw:
sure, just modified it.

william-silversmith:
Thank you Jingpeng!

zhihaozheng:
The new `np.all(point < self.maxpt)` would create a conflict with bbox created from other ways, e.g. [here](https://github.com/seung-lab/cloud-volume/blob/6ad71fc8c7412b9815f16a686ac27da469281a0a/cloudvolume/lib.py#L467), because it wouldn't pass [check_bounded](https://github.com/seung-lab/cloud-volume/blob/f4a90b776faf2fd67b8e5dd89cd0cc9a88136bfa/cloudvolume/datasource/precomputed/image/__init__.py#L65), right?

william-silversmith:
I don't think so. `from_list` just assigns the inputs you give it. If the inputs don't accord with acceptable parameters, then yes it will error.

zhihaozheng:
Igneous doesn't seem to follow this convention when creating bbox?

william-silversmith:
Are you running into an error rn?

zhihaozheng:
Yes I ran into a `OutOfBoundsError` that seems to come from `image.upload` during igneous downsampling, and the error is gone when I use the 6.1.1 version. But I haven't located which one in the process runs against this convention.

william-silversmith:
Can you post the full exception trace? It will help me locate and fix the error.

zhihaozheng:
```
 Traceback (most recent call last):
  File ".../lib/python3.8/site-packages/taskqueue/taskqueue.py", line 375, in poll
    task.execute(*execute_args, **execute_kwargs)
  File ".../lib/python3.8/site-packages/taskqueue/queueablefns.py", line 78, in execute
    self(*args, **kwargs)
  File ".../lib/python3.8/site-packages/taskqueue/queueablefns.py", line 87, in __call__
    return self.tofunc()()
  File ".../lib/python3.8/site-packages/igneous/tasks/image.py", line 456, in DownsampleTask
    return TransferTask(
  File ".../lib/python3.8/site-packages/igneous/tasks/image.py", line 433, in TransferTask
    downsample_and_upload(
  File ".../lib/python3.8/site-packages/igneous/tasks/image.py", line 81, in downsample_and_upload
    vol[new_bounds] = mipped
  File ".../lib/python3.8/site-packages/cloudvolume/frontends/precomputed.py", line 773, in __setitem__
    self.image.upload(img, bbox.minpt, self.mip, parallel=self.parallel)
  File ".../lib/python3.8/site-packages/cloudvolume/datasource/__init__.py", line 53, in guardfn
    return fn(self, *args, **kwargs)
  File ".../lib/python3.8/site-packages/cloudvolume/datasource/precomputed/image/__init__.py", line 199, in upload
    self.check_bounded(bbox, mip)
  File ".../lib/python3.8/site-packages/cloudvolume/datasource/precomputed/image/__init__.py", line 67, in check_bounded
    raise exceptions.OutOfBoundsError("""
cloudvolume.exceptions.OutOfBoundsError: 
        Requested cutout not contained within dataset bounds.

        Cloudpath: .....
        Requested: Bbox([32000, 0, 104],[64000, 32000, 105], dtype=int32)
        Bounds: Bbox([0, 0, 0],[64000, 64000, 300], dtype=int32)
        Mip: 1
        Resolution: [ 8  8 40]

        Set bounded=False to disable this warning.
```

william-silversmith:
That's very strange. That doesn't look like an off-by-one error. The requested bbox is clearly within the bounds. I'll have to do some experiments. Can you share with me what dataset this occurred with on slack?

j6k4m8:
↑ Also encountering this on 8.0.0, and 6.1.1 rollback fixed it. I can provide a MRE if you don't have one from @zhihaozheng above!

william-silversmith:
That would be fantastic Jordan! I'm working on another bug rn so this will help me a lot when I switch gears in an hour or two.

william-silversmith:
Actually, I just upgraded everything and ran the automated tests and saw the error. That should be sufficient! I should have run them on Igneous before releasing CV. Sorry about that!

j6k4m8:
Move Fast And Brain Things? :)

william-silversmith:
I reverted this change for now. I think Jingpeng's idea makes sense in isolation, but I now remember that I did this b/c it's much easier to make things line up b/c input and volume bounds match.

j6k4m8:
neato yeeto — thanks for the quick turnaround here!

PR #515, opened 2022-01-20
feat: add cv.unique

william-silversmith:
`cv.unique` allows you to query arbitrarily large bounding boxes and extracts the unique labels in an online fashion to avoid excessive memory consumption. 

I also exploit the cseg and compresso formats so that the extraction can occur after only the first stage of decompression for additional speed.


william-silversmith:
Parallel isn't supported in this PR, but it could be added later.

Issue #453, opened 2021-03-22
Passing tokens directly


bjm5164:
This check: https://github.com/seung-lab/cloud-volume/blob/master/cloudvolume/datasource/graphene/metadata.py#L124 prevents the ability to pass google auth tokens directly to a CloudVolume object. This has proven problematic for accessing our chunkgraph segmentation. Would it be possible to remove this check so that the `secrets` kwarg does not only accept hex tokens?

william-silversmith:
Hi bjm5164,

The issue with graphene is that it requires two kinds of tokens to access the underlying data: a graphene token (which communicates with the server) and the token for the storage service (s3 or GCS). I think we'd have to add another key for the graphene auth key in order to distinguish between them.

In the mean time, you can try using `CloudVolume(..., use_https=True)` to bypass the need for the secrets if your bucket is set to public-read.

bjm5164:
Hi Will,

Sorry, I think that I was incorrect in my wording. The token I want to pass is a graphene server token, but one not formatted as a hex string.

william-silversmith:
Sorry for the delay in getting back to you, I did some asking around. It looks like non-hex auth tokens are from alternative deployments of the graphene service. Up until now, I had no idea they existed. I talked to the author and they said they'll submit a PR with some small changes to help make CloudVolume compatible. Hopefully, this issue can be resolved soon.

As a workaround, you can put the token in a `chunkedgraph-secret.json` file which (for not any good reason) bypasses this check.

william-silversmith:
A fix was released in CloudVolume 3.9.0

PR #528, opened 2022-02-08
feat: move cave secrets from secrets to cave_secrets

william-silversmith:
This will allow specifying the storage secrets and the CAVE
secret at the same time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ You can create the `google-secret.json` file [here](https://console.cloud.google
 
 *Note: used to be called chunkedgraph-secret.json. This is still supported but deprecated.*
 
-If you have a token from Graphene/Chunkedgraph server, create the `cave-secret.json` file as shown in the example below. You may also pass the token to `CloudVolume(..., secrets=token)`.
+If you have a token from Graphene/Chunkedgraph server, create the `cave-secret.json` file as shown in the example below. You may also pass the token to `CloudVolume(..., cave_secret=token)` as a bare token string, json, or a dict containing the "token" field.
 
 ```json
 {
--- a/cloudvolume/cloudvolume.py
+++ b/cloudvolume/cloudvolume.py
@@ -72,7 +72,7 @@ class CloudVolume:
     background_color:int=0, green_threads:bool=False, use_https:bool=False,
     max_redirects:int=10, mesh_dir:Optional[str]=None, skel_dir:Optional[str]=None, 
     agglomerate:bool=False, secrets:SecretsType=None, 
-    spatial_index_db:Optional[str]=None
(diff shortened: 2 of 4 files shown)


fcollman:
can we do this is a deprecation warning kind of way and accumulate the breaking changes for a future major version change?

fcollman:
so for now as a patch change, add a deprecation warning that you should pass secrets in this other keyword argument, but still accept them in this one for graphene, then move the cleaner code version to a major version branch.

william-silversmith:
Yea that's no problem to keep it backwards compatible. I only considered doing it as breaking because there appear to be no references to secrets in PCG. I'll add in the shim.

william-silversmith:
It looks like the only code within the seung-lab github that would eventually need to be updated is this line: https://github.com/seung-lab/CAVEclient/blob/42e38a4fb8e0e73d8a911ca9edea18840f9ca786/caveclient/infoservice.py#L387

Not sure about other orgs though.

ceesem:
That code was just added (partially to deal with a case like this). Every time any code anywhere else is generating a cloudvolume — from both analysis scripts and services — it has to use the current secrets argument. The number of users who would be affected would span not only services but a significant fraction of people doing analysis or guidance scripting as well. This is a **very** significant breaking change and shouldn't be made lightly, even though I understand the hypothetical problem it solves.

ceesem:
For example, you could make a new parsing system that lets you specify secrets explicitly based on prefix, e.g.:
```python
cloudvolume.Cloudvolume(..., secrets={'graphene': cave_token, 'gs': google_secret})
```
that would resolve the ambiguity in the case you described but allow the current pattern to work while allowing the less verbose current nomenclature to persist during a period of deprecation.

william-silversmith:
Referencing issue https://github.com/seung-lab/cloud-volume/issues/453 

That makes sense to me and the design would accommodate growth. My goal is to always present as (seemingly) simple an interface as possible, so the old way will hopefully never be deprecated since that would impose a cost on every user since the most common use case is surely accessing raw precomputed from all kinds of datasets.

I think you're probably right that there are a few people using secrets as a parameter, but github search seems find few examples across all public codebases. My feeling is they probably mostly hiding in one-off scripts and Jupyter notebooks. 

https://github.com/search?l=Python&p=1&q=CloudVolume+secrets&type=Code

jefferis:
I haven’t thought through the implications of this change but just for completeness https://github.com/search?l=R&q=CloudVolume+secrets&type=Code.

william-silversmith:
Thought about this some more. I don't think Casey's suggestion will work because the secrets you input are already dicts. I think `cave_secret` (with backwards compatibility) is fine though perhaps a more generic name can be chosen to anticipate other use cases.

PR #529, opened 2022-02-11
feat: LRU cache for images

### LINKED ISSUE #526: LRU cache for images
@perlman had a great idea for accelerating certain kinds of lookup by adding an LRU cache. It would probably make sense for this cache to be per-instance rather than global (with some method to transfer the cache from instance to instance).

The example given was point lookups.

william-silversmith:
Resolves #526 

Repeated operations will be accelerated if they have spatial locality. By default the LRU is disabled.

```python
cv = CloudVolume(..., lru_bytes=1e8) # size in bytes
cv.image.lru.resize(1e9)
```
Note that if LRU and file caching are both enabled, the LRU can prevent some files from being written to cache if the cache is flushed without the LRU being flushed. As always with caches, if the data upstream changes without flushing the caches, update anomalies can occur.

- Adds `lru_bytes` argument to cloudvolume and `cv.image.lru` as a property.
- Adds thread safety to LRU
- Adds toggle between using number of elements and total byte size for LRU capacity (byte size not recursively calculated)
- fixes some hidden bugs on Windows where the chunk names might be using the incorrect slash type when listing chunknames in *some* callsites.
- fixes a bug in generating compressed morton codes when the input array is zero length.
- Adds sharding tests based on pinky40_v11 data and LRU tests


william-silversmith:
Things to remember to check:

1. Sharded
2. Caching
3. Reading after writing
4. Use of delete / exists

PR #538, opened 2022-04-27
 unnecessary integer casting for floating  chunk size

fcollman:
I discovered trying to work with a dataset with a floating point resolution that the draco mesh deduplication wasn't working correctly with graphene sharded.. i tracked the problem to this line, where casting the chunksize to integers caused the chunk boundary position prediction to accumulate error over many boundaries.... leading to incorrect detection of chunk boundaries and therefore improper stitching.

### DIFF
--- a/cloudvolume/datasource/graphene/mesh/sharded.py
+++ b/cloudvolume/datasource/graphene/mesh/sharded.py
@@ -235,7 +235,7 @@ class GrapheneShardedMeshSource(GrapheneUnshardedMeshSource):
     chunk_aligned_masks = []
     for mesh in meshes:
       level = meta.decode_layer_id(mesh.segid)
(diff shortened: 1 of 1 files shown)


william-silversmith:
Thank you for identifying, locating, and submitting the fix for this problem! 

In the future, I should probably write a helper function that does this to preserve the correct data type for the calculation:

```
    precision = max(map(lib.getprecision, fullres['resolution']))
    precision = max(precision, max(map(lib.getprecision, factor)))

    dtype = float
    if precision == 0:
      dtype = int
```

If this approach wouldn't provide further benefits for you, I'm happy to merge right now.

william-silversmith (approved):

fcollman:
I'm not sure if this section of the code it would provide much benefit..

PR #553, opened 2022-08-11
Reshape empty arrays to allow concatenation of empty skeletons

jasonkena:
Running `Skeleton.simple_merge(skels)` fails if any of the skeletons are empty, since the length of dimension 1 is set to 0. This change simply reshapes the empty arrays to the correct shape to correct for this.


william-silversmith:
Thank you Jason!!

PR #559, opened 2022-10-07
Pass secrets in PrecomputedImageSource.upload

perlman:
Secrets were not being passed to `tx.upload` on precomputed volumes.

I hit this as part of an experiment use cloud-files for https uploads (which will be a different PR). I found that info files, etc, were being created but the actual image data uploads were missing credentials.


william-silversmith:
Thank you Eric!

william-silversmith (approved):

PR #564, opened 2022-11-04
feat+perf(download_points): parallel point fetching

william-silversmith:
Precursor to fancy indexing.
image


william-silversmith:
What should the higher level interface look like?

There would be something of a natural fit with download_point but it would require making the function more comprehensive (accepting multiple points) and it would only improve things at size=1. I'd also have to detect with the input is a list of points vs a point, but generators make that difficult. I'm currently considering just going with `cv.scattered_points(pts)`.

PR #567, opened 2022-11-09
fix: ensure that at most max_mesh_layer will be requested from pcg

william-silversmith:


cloudvolume/datasource/graphene/metadata.py
```diff
@@ -402,8 +402,8 @@ def max_meshed_layer(self):
     The highest level in the ChunkedGraph that we create meshes for in this dataset.
     """
     if self.mesh_metadata and 'max_meshed_layer' in self.mesh_metadata:
```
akhileshh, line 404: Looks like `and 'max_meshed_layer' in self.mesh_metadata` is not needed anymore.

PR #603, opened 2023-12-05
Bugfix: fixes skeleton viewer

jasonkena:
Recommended way to instantiate Axes3D has led to breaking changes in matplotlib, causing axis to not render. See https://github.com/matplotlib/matplotlib/issues/24725#issuecomment-1351590000


william-silversmith:
Thank you! Can you add the recommended version for matplotlib in setup.py?

jasonkena:
Done! 3.6 was chosen as it seems to be the first version where the change happened.

william-silversmith:
Thank you!

PR #624, opened 2024-07-22
Fixes for lib.save_image

macrintr:
This is meant to fix two issues with lib.save_image:
1. The function requires the pillow library, but it's not included in requirements.txt.
2. There's a check if the image is np.bool that throws an error with the latest versions of numpy (I think the [np.bool alias expired with 1.24](https://numpy.org/doc/stable/release/1.24.0-notes.html)), causing an error when writing any dtype other than uint8.

### DIFF
--- a/cloudvolume/lib.py
+++ b/cloudvolume/lib.py
@@ -1065,7 +1065,7 @@ def save_images(
 
       if img2d.dtype == np.uint8:
         img2d = Image.fromarray(img2d, 'L')
-      elif img2d.dtype == np.bool:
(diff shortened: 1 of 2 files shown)


william-silversmith:
Thank you Tommy! This is the only place where Pillow is used in the code, would it be okay to move it to extras or to rewrite it using pyspng (used in chunks.py)?

macrintr:
Yeah, extras sounds good to me. Just made that change.

william-silversmith:
Thank you Tommy!

PR #625, opened 2024-07-27
feat: view function that can visualize images, seg, meshes, skeletons

william-silversmith:
```python
from cloudvolume import view

view([ image, skeleton, mesh ])
```

images will open microviewer.
meshes and skeletons will be co-visualized with vtk if installed.


william-silversmith:
This is integrated into microviewer as `microviewer.objects([ .... ])`.

https://github.com/seung-lab/microviewer

PR #628, opened 2024-08-02
feat: allow pointing directly at mesh and skeleton paths

william-silversmith:


william-silversmith:
This has already been implemented.

Issue #637, opened 2024-09-23
Allow passing of `segment_properties` argument to `CloudVolumePrecomputed.create_new_info`


chourroutm:
Hi,
The `CloudVolumePrecomputed` class already has `mesh` and `skeletons` arguments but is lacking `segment_properties` described in [the neuroglancer precomputed format docs](https://github.com/google/neuroglancer/blob/master/src/datasource/precomputed/volume.md#info-json-file-specification). This would be a named argument with default value to `None`, as for  `mesh` and `skeletons`.

PR #638, opened 2024-09-23
Add `segment_prop_dir` and `segment_properties` arguments for the precomputed format

chourroutm:
This PR adds a new argument to handle the "segment_properties" sidecar subdirectory and its info file.

References:
- https://github.com/google/neuroglancer/blob/master/src/datasource/precomputed/volume.md#info-json-file-specification
- https://github.com/google/neuroglancer/blob/master/src/datasource/precomputed/segment_properties.md

This will solve #637.


william-silversmith:
Hi! This is great, thank you! Is the PR ready for review? I was looking it over and it didn't seem to be actually doing anything with the `segment_properties` yet. Thanks again, this will be a great contribution!

chourroutm:
Hi! At the moment, it only adds the "segment_properties" field to the info file, to point to the subdir with the data. I'll have to write a CSV-to-segment_properties snippet anyway, which I can append to the PR. Should I withdraw the PR for now?

Also, I cannot `git clone` the files onto my Windows 11 workstation because of some filenames which contain ':'.

PR #658, opened 2025-04-07
Default to `encoding = 'raw'` when no compressor in Zarr

chourroutm:


chourroutm:
Thanks, @william-silversmith, I did not know that! I've edited the PR.

PR #659, opened 2025-04-09
refactor: drop posix_ipc and use native python shared_memory

william-silversmith:


william-silversmith:
This needs to be developed on a native linux machine.

PR #662, opened 2025-05-12
feat(zarr3): access zarr3 datasets

william-silversmith:


LucaMarconato:
Hi, amazing work! I have one comment on the file https://github.com/seung-lab/cloud-volume/blob/wms_zarr3/cloudvolume/datasource/zarr3/metadata.py. We are developing a library that takes care of handling metadata across various versions of the NGFF specs. It is still somehow work in progress but as it matures, it should simplify implementations (for instance of NGFF 0.6) and reduce maintenance burden.

So please keep an eye on it, or if you would like to already try it out (and give early feedback) you would be very welcome!

william-silversmith:
Hi, so I took a look at the package and in principle it looks pretty useful, but there is a potential issue. I am trying to keep the default set of dependencies small and this is an independent implementation of the zarr3 spec, even if it isn't the most comprehensive.

So, ideally there would be no dependency on the zarr3 reference package and the parser would accept zarr metadata to parse in the form of JSON or a python dict alone.

I do see a lot of value in using a metadata parser that can keep up with the evolving NGFF specification automatically.

Let me know if I am simply misunderstanding how to use it!

d-v-b:
> So, ideally there would be no dependency on the zarr3 reference package and the parser would accept zarr metadata to parse in the form of JSON or a python dict alone.

This makes perfect sense! Ddecoupling the metadata representation (i.e., a model of the spec) from the python implementation is a goal for `pydantic-zarr`, the library that `ome-zarr-py` uses, but it's a ways off (see [this branch](https://github.com/d-v-b/pydantic-zarr/tree/feat/optional-io)) and I don't have a timeline for when that will be complete.

d-v-b:
@william-silversmith since you did a fresh implementation of the spec, I'd also be interested in any feedback on what stuff we could have provided that would have made this job easier. Would a stand-alone package with just dataclasses and typeddicts for the core metadata models have been useful?

william-silversmith:
Yea, that would be useful so long as it came with a parser. It would have avoided some confusion I had. I was working from the specs, but also from example datasets and it took me some time to understand what was going on.

PR #665, opened 2025-06-10
Add more zfill to filename of 2D images

chourroutm:
the order of magnitude is rather 1000-10000 images than ~100


william-silversmith:
Thanks!

PR #666, opened 2025-06-11
feat(lib.save_images): Add support for little-endian 16-bit unsigned int

chourroutm:


william-silversmith:
Looks good to me!

https://pillow.readthedocs.io/en/stable/handbook/concepts.html#modes

chourroutm:
I got an error because of https://github.com/python-pillow/Pillow/blob/8ccdc399df1254c89bdb4e8fda6d6daf98943ab6/src/PIL/TiffImagePlugin.py#L1670-L1703:

> OSError: cannot write mode I;16L as TIFF

I'm correcting the PR!

PR #667, opened 2025-06-12
fix for "OSError: cannot write mode I;16L as TIFF"

chourroutm:


chourroutm:
Thanks! Ans sorry for the previous untested PR...

FYI I've tested it this time and it works

PR #676, opened 2025-09-26
fix(spatial_index): Replace StrEnum with str, Enum for Python 3.10 compatibility

### LINKED ISSUE #675: cloud-volume 12.6.1 fails on Python 3.10
gives error:

```
cannot import name 'StrEnum' from 'enum'
```

Google says this is a Python 3.11+ feature and I am on 3.10. Perhaps this usage could be avoided or a version pin could be added. Reverting to cloud volume 12.4.1 worked – We didn't try to track the change down further, Lisa and Greg Jefferis.

ranlu:
Fixes #675. Replaced StrEnum with str, Enum to support Python 3.10.

### DIFF
--- a/cloudvolume/datasource/precomputed/spatial_index.py
+++ b/cloudvolume/datasource/precomputed/spatial_index.py
@@ -7,7 +7,7 @@ import queue
 import sqlite3
 import threading
 import time
-from enum import StrEnum
+from enum import Enum
 
 import tenacity
 import numpy as np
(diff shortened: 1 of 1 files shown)


william-silversmith:
Would it make more sense to do something like this?

```try:
    from enum import StrEnum
except ImportError:
    from enum import Enum
    class StrEnum(str, Enum):
        def __str__(self):
            return str(self.value)```

ranlu:
Sure, why not, gemini says you do not need to define a __str__, because the class inherent from str already

PR #689, opened 2026-03-14
perf: Store numpy arrays directly in LRU cache instead of re-encoding

dodamih:
When lru_encoding is "raw", store decoded numpy arrays directly in the LRU cache instead of re-serializing them via encode_raw (which does np.asfortranarray + tobytes). This avoids significant overhead when reading non-"raw" data.

### DIFF
--- a/cloudvolume/chunks.py
+++ b/cloudvolume/chunks.py
@@ -151,7 +151,7 @@ def decode(
       return np.zeros(shape=shape, dtype=dtype, order="F")
     else:
       return np.full(shape=shape, fill_value=background_color, dtype=dtype, order="F")
-  elif isinstance(filedata, np.ndarray):
+  elif encoding == "numpy" or isinstance(filedata, np.ndarray):
     return filedata
   elif encoding == "raw":
     return decode_raw(filedata, shape=shape, dtype=dtype)
--- a/cloudvolume/chunks.py
+++ b/cloudvolume/chunks.py
@@ -151,7 +151,7 @@ def decode(
       return np.zeros(shape=shape, dtype=dtype, order="F")
     else:
       return np.full(shape=shape, fill_value=background_color, dtype=dtype, order="F")
-  elif encoding == "numpy" or isinstance(filedata, np.ndarray):
+  elif isinstance(filedata, np.ndarray):
     return filedata
   elif encoding == "raw":
     return decode_raw(filedata, shape=shape, dtype=dtype)
--- a/cloudvolume/datasource/precomputed/image/rx.py
+++ b/cloudvolume/datasource/precomputed/image/rx.py
@@ -496,27 +496,24 @@ def repopulate_lru_from_shm(
   ))
 
   encoding = lru_encoding
-  if encoding == "same":
-    # Since the parallel version populates the LRU via an image and
-    # you don't get the benefit of accessing the raw downloaded bytes,
-    # there will be a performance regression for "same" since e.g.
-    # jpeg -> img -> jpeg will instead of decode -> img,lru you'll
-    # have decode -> img -> encode -> lru. Therefore, this is hacky,
-    # but backwards compatible and strictly expands the capabilities
-    # of the LRU.
-    encoding = "raw" # would ordinarily be: meta.encoding(mip)
-
-  for chunkname in core_chunks[-lru.size:]:
-    bbx = Bbox.from_filename(chunkname)
-    bbx -= requested_bbox.minpt
-    img3d = renderbuffer[ bbx.to_slices() ]
-    binary = chunks.encode(
-      img3d, encoding,
-      meta.compressed_segmentation_block_size(mip),
-      compression_params=meta.compression_params(mip),
-      num_threads=meta.config.codec_threads,
-    )
-    lru[chunkname] = (encoding, binary)
+  if encoding in ("same", "raw"):
+    for chunkname in core_chunks[-lru.size:]:
+      bbx = Bbox.from_filename(chunkname)
+      bbx -= requested_bbox.minpt
+      img3d = np.copy(renderbuffer[ bbx.to_slices() ])
+      lru[chunkname] = ("numpy", img3d)
+  else:
+    for chunkname in core_chunks[-lru.size:]:
+      bbx = Bbox.from_filename(chunkname)
+      bbx -= requested_bbox.minpt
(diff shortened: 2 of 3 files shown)


cloudvolume/datasource/precomputed/image/rx.py
```diff
@@ -496,27 +496,24 @@ def repopulate_lru_from_shm(
...
-      meta.compressed_segmentation_block_size(mip),
-      compression_params=meta.compression_params(mip),
-      num_threads=meta.config.codec_threads,
-    )
-    lru[chunkname] = (encoding, binary)
+  if encoding in ("same", "raw"):
+    for chunkname in core_chunks[-lru.size:]:
+      bbx = Bbox.from_filename(chunkname)
+      bbx -= requested_bbox.minpt
+      img3d = np.copy(renderbuffer[ bbx.to_slices() ])
```
william-silversmith: Needs `order="F"`

william-silversmith (commented):
Thanks for taking a stab at this Dodam!

dodamih:
Addressed comments.

cloudvolume/chunks.py
```diff
@@ -151,7 +151,7 @@ def decode(
       return np.zeros(shape=shape, dtype=dtype, order="F")
     else:
       return np.full(shape=shape, fill_value=background_color, dtype=dtype, order="F")
-  elif isinstance(filedata, np.ndarray):
+  elif encoding == "numpy" or isinstance(filedata, np.ndarray):
```
william-silversmith: Hmm, seems worse than just testing if filedata is an ndarray. Maybe just remove  `encoding == "numpy"`. Sorry for the bad suggestion.

william-silversmith: Or if encoding is numpy and it's not a numpy array... throw an error?

dodamih: I think it's better to keep it simple and remove the encoding option check - "numpy" is fully internal anyway, and I think it's cleaner to just check the content. I'll push the change on top.

cloudvolume/datasource/precomputed/image/rx.py
```diff
@@ -608,35 +603,38 @@ def download_chunk(
...
-      fill_missing, mip, 
+      meta, filename, content,
+      fill_missing, mip,
       background_color=background_color,
       encoding=encoding,
     )
 
-  if lru is not None and full_decode:
-    if lru_encoding not in [ "same", encoding ]:
+  if lru is not None and lru_miss:
```
william-silversmith, line 618: This `lru_miss` logic is not quite right. Sometimes the LRU is populated with encoded data and not fully decoded but we lazily skip encoding it as the final encoding as we don't need it yet. For example, with `unique`, we sometimes can read the labels directly from the format without decoding. So if there is an LRU hit with the wrong encoding AND it is decoded to an image, we would like to re-encode it to the lru_encoding.

dodamih: Fixed.

cloudvolume/datasource/precomputed/image/rx.py
```diff
@@ -608,35 +603,40 @@ def download_chunk(
...
+    if full_decode and lru_encoding == "raw":
+      lru[filename] = ("numpy", img3d)
+    elif full_decode and lru_encoding not in [ "same", "raw", encoding ]:
```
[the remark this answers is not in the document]
dodamih: Resolved, but I'll note here that the 'lru_miss`' gate is gone above and moved down so we have to add 'raw' back here.

cloudvolume/datasource/precomputed/image/rx.py
```diff
@@ -608,35 +603,41 @@ def download_chunk(
...
       encoding=encoding,
     )
 
-  if lru is not None and full_decode:
-    if lru_encoding not in [ "same", encoding ]:
+  if lru is not None:
+    # Transcode on hit too: a prior call with full_decode=False may have
+    # populated the LRU with the original encoding, which we can now upgrade
+    # to lru_encoding now that we have a decoded image.
+    if full_decode and lru_encoding == "raw" and encoding != "numpy" and isinstance(img3d, np.ndarray):
```
william-silversmith, line 622: Shouldn't this read more like:

`if full_decode and lru_encoding in ("raw", "numpy") and isinstance(img3d, np.ndarray):`

cloudvolume/datasource/precomputed/image/rx.py
```diff
@@ -608,35 +603,41 @@ def download_chunk(
...
+      lru[filename] = ("numpy", img3d)
+    elif full_decode and lru_encoding not in [ "same", "raw", encoding ]:
```
william-silversmith, line 624: What if `lru_encoding` is `"raw"` or `encoding` and there is an lru hit?

cloudvolume/datasource/precomputed/image/rx.py
dodamih: Not as is - note that `lru_encoding` (the user facing option) cannot be "numpy"; we're changing the "raw" `lru_encoding` option to use `encoding` (the non-user facing option) "numpy", but now we have to handle situations where the `encoding` is "raw" (meaning bytes) or some other compression scheme with a LRU hit that requires deserialisation into numpy.

cloudvolume/datasource/precomputed/image/rx.py
dodamih: The logic as it is handles it, right?

`lru_encoding` is `"raw"`: if internal encoding is not `"numpy"`, caught by first branch; if it is, we're done.
`lru_encoding` is `encoding`: if internal encoding is `encoding`, this doesn't fire (as it should not); if internal encoding is different, then this fires and transcodes to the user-desired `lru_encoding`.

cloudvolume/datasource/precomputed/image/rx.py
william-silversmith: Oh I see, because when it is retrieved from LRU, `encoding="numpy"`. You want to avoid an extra assignment when it is redundant.

cloudvolume/datasource/precomputed/image/rx.py
william-silversmith: Ok this is making sense to me now.

william-silversmith:
Thank you for sticking with this Dodam! <3

PR #692, opened 2026-04-15
perf: faster all label query with psql

ranlu:
It turned out the fastest approach to query all the unique labels with psql is to dump all the labels then dedup with unique, it reduces processing a data containing 2B unique labels from 4 hours to 15 minutes. I think this is because the labels are mostly unique, only the very large segments can span multiple chunks, so it is not efficient to use complex database query to jump over redundent labels.

### DIFF
--- a/cloudvolume/datasource/precomputed/spatial_index.py
+++ b/cloudvolume/datasource/precomputed/spatial_index.py
@@ -842,17 +842,13 @@ class SpatialIndex(object):
     conn.close()
     return locations      
 
-  def query(self, bbox, allow_missing=False, nthread=1):
+  def query(self, bbox, allow_missing=False):
     """
     For the specified bounding box (or equivalent representation),
     list all segment ids enclosed within it.
 
     If allow_missing is set, then don't raise an error if an index
     file is missing.
-
-    nthread: number of threads to use for the Postgres fast path
-    (when the query covers the full dataset). Has no effect for
-    non-Postgres databases or non-fast-path queries.
 
     Returns: iterable
     """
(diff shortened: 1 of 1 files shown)


cloudvolume/datasource/precomputed/spatial_index.py
```diff
@@ -137,57 +127,32 @@ def _build_pg_binary_copy_two_bigints(col1, col2):
...
-      for i in range(actual_threads)
-    ]
-    arrays = [f.result() for f in futures]
-
-  non_empty = [a for a in arrays if len(a) > 0]
-  if not non_empty:
-    return np.array([], dtype=np.uint64)
-
-  return np.concatenate(non_empty)
+  labels = fastremap.unique(labels)
```
william-silversmith, line 153: Shouldn't the results already be sorted?

ranlu: Yes in current implementation. I suppose this is a semantic thing. fastremap.unique interface does not indicate the result should be sorted, so I do not want to depending on that, we might have faster implementation in the future that does not sort the results.
On the other hand, does query requires the final results to be sorted? I did this because the mysql path sort the results, but the json branch do not sort the results.

william-silversmith: So good news for you, the branch of fastremap.unique that I'm working on supports the `sorted` flag and has the kind of behavior you are referencing for the very specific combination of `sorted=False + uint64 + [N,2]`

https://github.com/seung-lab/fastremap/pull/58

PR #701, opened 2026-08-11
fix: import toiter in cacheservice

nkemnitz:
`CacheService.delete()` calls `toiter` but it was never added to the `from .lib import (...)` list, so every call raises `NameError`. Been that way since the method was added in #560 (Oct 2022).

Only in-tree caller is the unsharded mesh delete, where `cf.delete(filenames)` runs first and `self.cache.delete(filenames)` then raises — so deleting meshes with a cache enabled fails after the remote objects are already gone.

```python
cv = CloudVolume(path, cache=cache_dir)
cv[bbox]
cv.image.cache.delete(['4_4_40/0-64_0-64_0-64'])
# NameError: name 'toiter' is not defined
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)


william-silversmith:
Thank you!

PR #703, opened 2026-08-13
perf: wait on condition variable instead of poll sleep

william-silversmith:


nkemnitz:
This one can silently drop jobs if the generator is a bit slow, for whatever reason:

```python
import time
from cloudvolume.scheduler import schedule_threaded_jobs

def fns():
  for i in range(50):
    time.sleep(0.001)      # producer slower than the 20 consumers
    yield lambda i=i: i

got = schedule_threaded_jobs(fns(), concurrency=20)
print(f"{len(got)}/50 jobs ran")

# master   50/50 jobs ran
# pr703    15/50 jobs ran
```

Also if the first thread runs out of tasks, it will set the shared terminate event for all other threads. And any Exception raised by the still in-flight tasks won't get caught anymore, I believe?

